PR #1, opened 2019-08-12
Initial setup

alberts2:
Initial setup of the C2C repo based on [svn10983](https://sourceforge.net/p/caveman2cosmos/code/10983/) without any wonder or other movies.


billw2012:
I guess we may as well stay in this branch until we are ready for everyone to move over right? i.e. rest of the stuff implemented like FPK live, appveyor build creation, version.h, source indexing.

alberts2:
I didn't include the *.bik  movies because:

- They use alot of space.
- Manny of them are just static images.
- We don't have movies for everything and images in the dds format are easier to create.
- Some of them have a very low quality.

alberts2:
**No longer compatible with older saves!!**
The reason for this is that the commit sha hash is now stored in saves and used to trigger the recalc if the hash changed since a game was saved insted of using the svnRevision.

alberts2:
@billw2012 @Thunderbrd 
I made a few changes which should keep save compatibility with the svn repo but I didn't have much time to test it.

alberts2:
I think there's a error in the save read or write code changes i made in this branch but i'am not yet sure where?

PR #19, opened 2019-08-20
Use the Appveyor build Cache. 

alberts2:
Using the build Cache to speed up the dll compilation.


billw2012:
Although it isn't really obvious to me that the cache will be faster than unzipping the deps.exe locally.

alberts2:
That's only the first step the next step is adding Sources\temp_files\Debug and build the debug dll insted of rebuilding it completely each time.

alberts2:
The problem i'am having with this is that fastdep uses the file modification date but git doesn't file timestamps.

alberts2:
Okay the fix_sources_timestamps.ps1 sems to work now.

alberts2:
Xerces was compiled with /GL if we don't use /LTCG in all configurations it will turn on by itself.

`XercesLib-3.1.1-x86-windows-vc-7.1.lib(AbstractDOMParser.obj) : warning LNK4218: non-native module found; restarting link with /LTCG
   Creating library temp_files\Debug\CvGameCoreDLL.lib and object temp_files\Debug\CvGameCoreDLL.exp
Generating code
Finished generating code`

billw2012:
So if I am correct then `/LTCG` has no effect except for objs compiled with `/GL` therefore having it enabled for all builds still only affects FinalRelease?

alberts2:
Xerces was compiled with `/GL` it will always be linked with `\LTCG` enabled in all configurations. But the CvGameCore source code is only in the FinalRelease configuration compiled with `\GL`.

alberts2:
AppVeyor seems to have a bug when a build Cache is invalidated.

`Build started
git clone -q --branch=appveyor-build-cache https://github.com/caveman2cosmos/Caveman2Cosmos.git C:\projects\caveman2cosmos
git checkout -qf eb4c036b2a34e9e22281821a0471c2423ceaeb7f
Restoring build cache
Cache 'Sources\temp_files\Debug' - Invalidated as some dependencies (Sources\deps, Sources\makefile, appveyor.yml) have changed
Cache 'Sources\temp_files\Release' - Invalidated as some dependencies (Sources\deps, Sources\makefile, appveyor.yml) have changed
call Powershell.exe -NoProfile -ExecutionPolicy Bypass -File "%APPVEYOR_BUILD_FOLDER%\Tools\CI\fix_sources_timestamps.ps1"
call "%APPVEYOR_BUILD_FOLDER%\Tools\_MakeDLL.bat" build Debug
Checking dependencies...
Dependencies not found! 
Unpacking Caveman2Cosmos DLL dependencies, please wait for the process to complete...`
Unpacking the deps.exe hangs when a cache is invalidated.

billw2012:
Yeah perhaps it is popping up a dialog asking if user wants to overwrite existing files or something. I changed the logic so that it checks for nmake.exe specifically instead of just the existence of the deps folder, because I kept getting empty deps folder left behind when changing branches. 
So we just need to add a `rmdir /Q /S deps` (maybe the flags aren't right) command into the InstallDeps.bat file.

alberts2:
I made an error with the cache dependency settings but the deps.exe waiting for user input could also explain some things.
But first i try if fixing the dependencies helps.
Caching the Debug and Release *.obj files cuts the appveyor run time almost in half.

PR #50, opened 2019-08-25
Including drd_avel's reworked gamefonts

KaTiON-PT:
210+210 icons were remastered


KaTiON-PT:
I think I did something wrong

PR #52, opened 2019-08-25
Prehistoric Era Revamp

KaTiON-PT:
Balance changes to tech ordering, buildings, localization and units.


raxo2222:
I noticed, that some modules, that should be deleted weren't deleted.
Also I'm not sure if updating branch to latest master version was needed, but I was testing it anyway.

Thunderbrd:
So... does someone have to merge this pull request?  Are we able to do that now?

PR #61, opened 2019-08-28
Exposing functions for mouse position to python

billw2012:


Thunderbrd:
POINT... interesting function.  Not familiar with that.  What does it do?  Send the screen location of the pointer or something?

PR #72, opened 2019-09-01
Initial camera zoom level and FOV

Toffer90:
Changed default FoV from 50 to 42.
Tweaked how initial zoom level is set a bit more.


Toffer90:
This is final. I've nothing more to add in regards to the music transition by zoom level issue.

PR #73, opened 2019-09-01
Ignore *.nif files in Assets

alberts2:
All `*.nif` files in the `Assets` directory should be ignored because they should be placed in `UnpackedArt` instead.


alberts2:
If anybody agrees with this change please Accept this Pull Request.

billw2012:
Was adding `/Build` back intentional?

alberts2:
I think that's just some kind of whitespace or line ending change.

PR #92, opened 2019-09-03
SVN rev. 11027

Toffer90:
War weariness calculation fix
Ranged bombard mission fix
UI tweaks
Better Map bonus buttons
Misc fixes


billw2012:
Yeah I meant to say I want to push an update, and you beat me to it! I'm not a fan of setting the precedent of putting svn rev number in this though, I think it is only incidental at this point. Personally I would prefer it highlight a specific topline feature we are putting in.

I can automatically tag the commit with the SVN revision number if we want to do that though, I think it might be a good idea for translating bugs reported in SVN back here.

Toffer90:
If you figure out the protocol on these things, great.
I'll be happy to follow it, I'm just improvising those names and comments in the absence of a recipe.

PR #93, opened 2019-09-04
Value to units for buildings constructable throughout nation

Thunderbrd:
See what y'all think.  I'm not quite sure if this is slower than it should be still or not but it should be as streamlined as the concept can get and we have a new global to adjust for the assumable value of a building.  This is crude but may work as a result.


billw2012:
So for checking perf you should run a few EOT on a late game save without these changes and time them, then do the same with these changes and time them (starting from the same save game, so it's the same turns), then compare the results.
Anything more than 5% or so change would need to be looked into further I think.
Also you can see the errors you are getting are in the FAssertMsg calls. These are only enabled in Debug and Assert builds, so I guess you only tested this in Release.

Sources/CvPlayerAI.cpp
```diff
@@ -12852,9 +12852,58 @@ int CvPlayerAI::AI_unitPropertyValue(UnitTypes eUnit, PropertyTypes eProperty) c
...
+		{
+			BuildingTypes eBuilding = (BuildingTypes)kUnitInfo.getBuildings(iI);
+
+			if (NO_BUILDING != eBuilding)
+			{
+				if (canConstruct(eBuilding, false, false, true) && AI_getNumBuildingsNeeded(eBuilding, bCoastal) > 0)
+				{
+					iCount = 0;
+					iBuildingValue = 0;
+					for (CvCity* pLoopCity = firstCity(&iLoop); pLoopCity != NULL; pLoopCity = nextCity(&iLoop))
```
billw2012: I bet looping over city is slower than looping on the outer loop. You could instead do:
```
foreach city
    if city is costal etc
        for each unit building { ... }
```
i.e. Invert the loop so that you only do the slow loop once.

Sources/CvPlayerAI.cpp
```diff
@@ -12852,9 +12852,58 @@ int CvPlayerAI::AI_unitPropertyValue(UnitTypes eUnit, PropertyTypes eProperty) c
...
+		if (bCoastal)
+		{
+			iMinOceanSize = GC.getMIN_WATER_SIZE_FOR_OCEAN();
+		}
+
+		for (int iI = 0; iI < kUnitInfo.getNumBuildings(); iI++)
+		{
+			BuildingTypes eBuilding = (BuildingTypes)kUnitInfo.getBuildings(iI);
+
+			if (NO_BUILDING != eBuilding)
```
billw2012: Is it ever actually possible for this to happen? Surely if you are looping over the buildings of the unit then they will all be valid buildings. If this is the case then this check should be an `FAssertMsg` instead like `FAssertMsg(eBuilding != NO_BUILDING, "UnitInfo buildings must be valid");`.
This improves readability, and performance, and better expresses expectations of the data.

Sources/CvPlayerAI.cpp
```diff
@@ -12852,9 +12852,58 @@ int CvPlayerAI::AI_unitPropertyValue(UnitTypes eUnit, PropertyTypes eProperty) c
...
+			{
+				if (canConstruct(eBuilding, false, false, true) && AI_getNumBuildingsNeeded(eBuilding, bCoastal) > 0)
+				{
+					iCount = 0;
```
billw2012: Why are we resetting iCount here? It means we are only counting the number of places we can build whatever happens to be last in the units building list. This makes no sense!

Sources/CvPlayerAI.cpp
```diff
@@ -12852,9 +12852,58 @@ int CvPlayerAI::AI_unitPropertyValue(UnitTypes eUnit, PropertyTypes eProperty) c
...
+					iBuildingValue = 0;
+					for (CvCity* pLoopCity = firstCity(&iLoop); pLoopCity != NULL; pLoopCity = nextCity(&iLoop))
+					{
+						if ((!bCoastal || pLoopCity->isCoastal(iMinOceanSize)) && (pLoopCity->area() == pArea))
+						{
+							if (pLoopCity->getNumBuilding(eBuilding) == 0 && pLoopCity->canConstruct(eBuilding, false, false, true))
```
billw2012: These can be collapsed into a single if statement instead of two nested ones. The || and && operators only evaluate their right hand side if they need to. 

e.g.
`if (true || expensiveFunction()) {}` will NEVER call `expensiveFunction()` because it can't change the result of the expression. Regardless of what `expensiveFunction()` returns the if statement would always be executed, because `true || true` is `true` and `true || false` is also `true`.

`if(false && expensiveFunction()) {} `will NEVER call `expensiveFunction()` because it also can't change the result. Regardless of what it returns the if statement will never be executed because `false && true` is `false` and `false && false` is also `false`.

The benefits of reducing it to a single statement are readability (deep nesting reduces readability) and expressiveness: 
It accurately expresses that you require all these conditions to be met to increase iCount, and that you are *not* implying that there could be other possibilities when the outer if statement is true but the inner one is false. i.e. It tells future programmers that they should not be considering adding an else statement to the inner if statement to handle where it fails.

Sources/CvPlayerAI.cpp
```diff
@@ -14395,15 +14498,15 @@ int CvPlayerAI::AI_unitValue(UnitTypes eUnit, UnitAITypes eUnitAI, CvArea* pArea
...
+
 	if ((iCombatValue > 0) && ((eUnitAI == UNITAI_ATTACK) || (eUnitAI == UNITAI_ATTACK_CITY)))
 	{
 		if (pArea != NULL)
 		{
 			AreaAITypes eAreaAI = pArea->getAreaAIType(getTeam());
 			if (eAreaAI == AREAAI_ASSAULT || eAreaAI == AREAAI_ASSAULT_MASSING)
 			{
-				for (int iI = 0; iI < GC.getNumPromotionInfos(); iI++)
+				for (iI = 0; iI < GC.getNumPromotionInfos(); iI++)
```
billw2012: This is wrong, variable should always be defined it the scope where it is used. For loop index this means in the loop variable as it was originally.

Sources/CvPlayerAI.cpp
```diff
@@ -12852,9 +12852,58 @@ int CvPlayerAI::AI_unitPropertyValue(UnitTypes eUnit, PropertyTypes eProperty) c
 	return iValue;
 }
 
-/********************************************************************************/
-/* 	City Defenders						24.07.2010				Fuyu			*/
-/********************************************************************************/
+int CvPlayerAI::AI_unitBuildingValue(UnitTypes eUnit, CvArea* pArea) const
+{
+	FAssertMsg(eUnit != NO_UNIT, "Unit is not assigned a valid value");
+	FAssertMsg(eUnitAI != NO_UNITAI, "UnitAI is not assigned a valid value");
```
billw2012: There is no eUnitAI, this line can just be removed. It appears it is valid for pArea to be null (this should be noted in a comment on the function, or by giving a default parameter of null in the declaration). Although its kind of weird for this function to just return 0 if pArea is null. When is that useful? Instead it should either evalute the building for all areas if area is null, OR it should just not be called at all, and require area to not be null.

Sources/CvPlayerAI.cpp
Thunderbrd: This was a conversion artifact when removing the buildingclass reference where it WAS possible for eBuilding to refer to NO_BUILDING in the case of some bad XML.  I caught one of them but this is a good spot.  Is this even a current project still though?  I don't know where we are at with all this.  Toffer did something completely altered didn't he?

Sources/CvPlayerAI.cpp
Thunderbrd: Hah... makes sense and a great observation.  I'll look for that in the future.  My thinking was why loop the cities if you don't have to but you WILL have to at least once so I'm setting us up to loop through them more than that rather than to avoid having to at all.  Right.

Sources/CvPlayerAI.cpp
Thunderbrd: hmm... I think the thinking on this is that the first set of checks relate to each other and you could put a stop at each layer of related checks to more easily determine where a check set is failing if you feel it should be succeeding or succeeding where you feel it should be failing or simply helping to determine a smaller segment of possible reasons for failure of reaching the final result to a more limited selection of checks by not trying to stuff all the checks into one IF statement line.  So I'm not sure that I agree that deep nesting reduces readability - in fact I feel its very much the opposite in that it makes it even harder to read and harder to narrow in on a culprit spot during debug analysis.
It might come down to what you're used to seeing I suppose.  Is there a distinction much in the processing efficiency?

Mind you, I didn't invent this method of multiple nesting of if statements but I've found it very helpful when debugging and figured I should mention how.

Sources/CvPlayerAI.cpp
Thunderbrd: Must admit I avoid setting up FAsserts because I figure I'll get myself confused on whether I should be expressing a statement that I expect to be true or false.  But this was, as mentioned, an unnecessary artifact from when this was originally a tag expressing a buildingclass rather than a direct building type reference.

Sources/CvPlayerAI.cpp
Thunderbrd: Please clarify as there has been some differences of opinion among various programmers on this matter:
is (int iI = 0... or (iI = 0.... (with the int iI being declared at the beginning of the function instead of within the for statement) correct.  I'm not sure what you mean by defined in the scope where it is used.  My understanding was that it should be (int iI = 0...  But this is very much not the standard that Firaxis started with, which was to declare iI at the beginning of the function usually.  I don't usually fight that if that's how the function is when I write a For loop.

Is there a reason for one to be considered better form than the other and what are the advantages/disadvantages?  (can't promise I'll understand the answer but I'll try.)

Thunderbrd:
Is this still an active project and can someone help me understand where we are with it?  I saw somewhere Toffer saying that he had a whole 'nother proposal about how to go about this.  Should we be starting over with that proposal instead?

Toffer90:
I consider it an active project still.
The suggestion I made is in a new branch, from this branch, called unitAiValuationAlt.
The commits made to unit-ai-worker (this) branch by raxo after the branch point for unitAiValuationAlt are unrelated to these branches and can be ignored.

Thunderbrd:
1) How would one 'ignore' a commit?  Just unwrite it with a new one?  Or revert it out somehow?
2) So I should be looking at unitAIValuationAlt instead of this branch anyhow?  Does that mean this one gets abandoned entirely?  Are we leaving it undeleted for now for the sake of reference?  There are some damn good training notes in this thread...

Toffer90:
> 1. How would one 'ignore' a commit?  Just unwrite it with a new one?  Or revert it out somehow?
> 2. So I should be looking at unitAIValuationAlt instead of this branch anyhow?  Does that mean this one gets abandoned entirely?  Are we leaving it undeleted for now for the sake of reference?  There are some damn good training notes in this thread...

By ignore I mean there's no need to inspect what changed in those commits as the changes have nothing to do with what this branch is about. We could continue with the alternative branch if you, or other dll modders, think the suggestion is ok.

Thunderbrd:
OK, I'll try to look into the alternative branch very soon.

Sources/CvPlayerAI.cpp
billw2012: > So I'm not sure that I agree that deep nesting reduces readability

It's pretty much 100% agreed on in the programming community at large. [Here](https://en.wikibooks.org/wiki/Computer_Programming/Coding_Style/Minimize_nesting) is one of the many explanations you can find. 
This specific case it isn't so bad, as it all fits on a few lines. However I have seen plenty of places in the code where if statements are nested 3+ deep, and this is well understood to be bad for readability and maintainability.

> Is there a distinction much in the processing efficiency?

There is possibly a distinction in how well the optimizer can analyse it. Generally the more semantically concise your code is the better the optimizer can reason about it. This goes for people reading the code as well. If you see a single if statement you know that it models a single condition that the inner block relies on. If you see nested if statements you now need to consider the context of each one. Does it have an else statement, does it have code that is executed for one of the outer blocks but not the inner one? It is a less precise way of expressing your intent, and requires the reader and optimizer to look at all the code before they can conclude that you actually just intend this to be a single condition with a single piece of code executed if it passes.

> Mind you, I didn't invent this method of multiple nesting of if statements but I've found it very helpful when debugging and figured I should mention how.

Nobody invented it, it's just a typical naive coding style. i.e. It's something you learn is bad from having read and maintained a lot of code bases in various styles. Most coders will, at a glance, tell you that this code base is poorly written, and the deep nesting of statements will probably be the thing they are basing that on.

The "autos" window will help you debug if statements, it shows the return values of all functions and useful variables so you can see how an if statement was evaluated. 
Another option with complicated if statements is to consider encapsulating the condition in a function. This can make the code a lot more readable because the function name simply describes the condition you are testing, and the reader doesn't need to know what exactly the function does.

Sources/CvPlayerAI.cpp
billw2012: > Please clarify as there has been some differences of opinion among various programmers on this matter:
Well you know what they say about opinions! If you want to know the best practice on something you should trust google/stackoverflow/expert blog posts (there is a large array of experts in C++ who write on these kind of things, e.g. Scott Meyers), over the opinions of any individual programmer.
When I talk about these things related to style, structure etc., I'm relaying the information I learned from these sources, and confirmed in my own experience as being good. i.e. You can confirm the things I'm saying with google, and where you can't I will be happy to update my understanding of best practices. 

> is (int iI = 0... or (iI = 0.... (with the int iI being declared at the beginning of the function instead of within the for statement) correct. 
The first is always correct. You should never be defining variables at the start of a function, this is a holdover from very old C style programming when it was required to declare variables at the start of a scope (C89 required this, but not C99, that is how old we are talking). Lots of C++ programmers in the early days were C programmers who didn't bother to learn how to write C++ correctly, because they thought it was basically the same thing but with classes, or they preferred to stick with the style they knew.

> I'm not sure what you mean by defined in the scope where it is used.
A scope in C++ is basically everything `{` between two matching curly braces `}`. Other structures can define a scope, but that is the main one in functions. So in a for loop the scope of the for loop is the `{ ... }` that is executed for each iteration.
I guess I actually meant to say "declared in the scope where it is used.", rather than defined. In other words if you are using a variable inside a for loop, and it's only used there, you should declare it there as well.
i.e. Instead of:
```cpp
int i, value;
for(i = 0; i < 10; i++)
{
    value = 0;
    value += get_value(i);
    if(value > 5) 
    {
        max_value = value;
        // etc
    }
}
```
You should write:
```cpp
for(int i = 0; i < 10; i++)
{
    int value = get_value(i);
    if(value > 5) 
    {
        max_value = value;
        // etc
    }
}
```

> Is there a reason for one to be considered better form than the other and what are the advantages/disadvantages?

As I alluded to above, I'm not giving my random opinions anywhere here when it comes to what is good coding practice or style (when I do I will be clear that it is just my opinion rather than established best practices), these are conventions of good software engineering, and are well supported by reasoning and experience of the community as a whole.

One reason I can give, that is particularly salient in our case, is that I already found a few bugs in the code that were caused due to declaring variables at the start of a function instead of where they are used. For instance in one case a variable was used in a for loop (and of course declared at the start of the function), but the coder forgot to reset the value at the start of each loop, so it ended up corrupting the final calculation. If they had declared the variable at the start of the loop then the compiler would have warned them if they didn't initialize it before using it. In other cases the same variable gets reused in different places, but the person forgot to reset it, so the value in the second location is corrupted by the value in the first. In this case it is *two* bad coding practices, firstly declaring the variable before it is used, and secondly reusing a variable for two different purposes.

Another reason is readability. Readers expect variables to be declared in the scope in which they are used, so seeing a variable being declared indicates that it is used in that scope. It is much easier to understand the use of a variable when it is declared where it is used. i.e. You know it is only used in that scope. Another way to phrase this (which applies to a lot of style guidelines), is that code should clearly (and unambiguously, and concisely) indicate the intent of the programmer. 

Performance is another reason: declaring a variable is allocating it on the stack. That is fast, but it isn't free. If you declare a bunch of variables and then check an if statement and leave the function before using them, you wasted time allocating them.

And again, it confuses the optimizer. It wants to optimize where variables are allocated, what registers they are assigned to etc., and this is much easier for it when it can easily determine what section of the code the variable is actually used in. Declaring all the variables at the top of the function is you telling the optimizer that you could be using these variables anywhere.

billw2012: > But this is very much not the standard that Firaxis started with, which was to declare iI at the beginning of the function usually. I don't usually fight that if that's how the function is when I write a For loop.

I mentioned already I guess, but I would not use anything in this code base as an example of how to write good code. Its not even a matter of just using the existing code style in the case of all other things being equal. In this case continuing to use the existing code style is actually bad for the health of the code and the future maintainability. It's better to start any new code or functions using correct style and practices, and to update existing functions as and when we can. e.g. I will update any function I am significantly modifying. I already fixed large portions of CvCity and CvUnit classes (not all though).

Also a further note on style and correctness in C++: there are many professional high quality code bases available for browsing on github. You can delve into any of them to see various examples of what good code should look like. e.g. I tried just now scrolling through a few files in the Unreal code base (which has some problems, but is pretty good generally) to look for if nesting. I couldn't find any more than 2 deep, and the ones that *were* 2 deep without else statements were done for optimization purposes, e.g. where the result from the inner if statement is used in the the inner block, like:
```cpp
if(a > b) 
{
  Ptr* obj = get_best_ptr(a, b);
  if(obj != nullptr)
  {
    obj->do_a_thing();
  }
}
```
However even in this case it is questionable. Perhaps `get_best_ptr` and the following if statement should instead by replaced by `try_do_a_thing_to_base_ptr(a, b);`.

PR #139, opened 2019-09-09
Add per object memory tracking replacing existing system

billw2012:
It doesn't show leaks, however the old system would only show them for a hand full of objects anyway, this new system can show them all if we want to.


alberts2:
Would it be possible to track the parent object which e.g. allocated a `int` as well?

billw2012:
> Would it be possible to track the parent object which e.g. allocated a int as well?

Well it determines the object just by making a template operator function that then uses `typeid()` on the pointer type so it knows nothing about who owns that pointer. Of course we can work out how to do it, however the allocations also all refer to the allocation location anyway. I was going to make a report that just lists file and line ordered by allocation total which should make it clear. However after discovering we only allocate about 25% of the memory used in the DLL vs 75% in the EXE I didn't bother to extend it further.

alberts2:
> However after discovering we only allocate about 25% of the memory used in the DLL vs 75% in the EXE I didn't bother to extend it further.

I kind of expected a outcome like that from a few huge memory leaks we had in the past it was obvious that the exe uses most of the memory. But it's good to have proof of that suspicion.

PR #142, opened 2019-09-10
Changed tech names to be better representative (pedia entries as well)

GreatCoffers:


raxo2222:
Why you committed deleting file on your fork? All you had to do was reverting it, and redoing thing without changing tabs to spaces - use notepad++ to edit file.

PR #143, opened 2019-09-10
Tech name and pedia changes.

GreatCoffers:


raxo2222:
> 
> 
> Am I correct in understanding that you are deleting the old tech_civ4gametext.xml and adding an altered one in its place? That makes it VERY hard to compare the two. I'm still learning Git here so I'm not sure I'm seeing this correctly.

No - he made mistake, and accidentally converted tabs to spaces.
He needs to redo stuff again.

KaTiON-PT:
@Thunderbrd here are his [changes](https://docs.google.com/document/d/1hQeWf8sAG5_lbmsMNreaa9SlSVhhXwo-MuKW6yeb1ss).

I'm of the opinion that there a few of these changes are unnecessary as they would step on the toes of pepper2000.

Some don't really make sense (like Genesis Biology -> Phylogenetics) Phylogenetics is considered a subset of a subset of biology that has existed since the [19th century](https://en.wikipedia.org/wiki/Ernst_Haeckel) and what Genesis Biology is going for is unlocking the secrets of life, so i think it should still be called that.

GreatCoffers:
> @Thunderbrd here are his [changes](https://docs.google.com/document/d/1hQeWf8sAG5_lbmsMNreaa9SlSVhhXwo-MuKW6yeb1ss).
> 
> I'm of the opinion that a few of these changes are unnecessary as they would step on the toes of pepper2000.
> 
> Some don't really make sense (like Genesis Biology -> Phylogenetics). Phylogenetics is considered a subset of a subset of biology and has existed since the [19th century](https://en.wikipedia.org/wiki/Ernst_Haeckel) and what Genesis Biology is going for is unlocking the secrets of life, so I think it should still be called that. If what GreatCoffers is going for is modern phylogenetics then it is 100% included in the already existing Bioinformatics tech.

I've talked with pepper2000. He's ok with these renames. [PMs](https://forums.civfanatics.com/conversations/tech-name-changes.496099/)

Genesis Biology sounded dumb to me. It sounds too religious and its pedia is talking about Phylogenetics, so that's why it was changed. 

[Bioinformatics](https://en.wikipedia.org/wiki/Bioinformatics) is the field where computers are developed to understand and interpret biological data.  [Phylogenetics ](https://en.wikipedia.org/wiki/Phylogenetics) is the study of relationships between organisms.  Bioinformatics is the computerization of biological understanding, while Phylogenetics is the way we understand relationships between organisms, so that we may resurrection, them, etc.

PR #144, opened 2019-09-12
Value to units for buildings constructable throughout nation

alberts2:
Newer version of #93


Toffer90:
> I resolved the conflict in A_New_Dawn_GlobalDefines.xml and updated this branch.
> 
> Are there any other known issues with these changes?

There was a performance concern, not sure if that was resolved adequately or not.
Someone should profile it and find out.

raxo2222:
It seems like it fell out.

Toffer90:
> It seems like it fell out.
> 
> Its so old, that there are two conflicted files now

Wasn't very important anyhow, the conflicts should be easy to resolve, as this branch did very little and should be easy to understand if one compare it to the master version it branch out from where there's no conflicts.

Sources/CvCityAI.cpp
```diff
@@ -4858,198 +4848,161 @@ UnitTypes CvCityAI::AI_bestUnitAI(UnitAITypes eUnitAI, int& iBestValue, bool bAs
...
 	}
 
-	CvCivilizationInfo& kCivilizationInfo = GC.getCivilizationInfo(getCivilizationType());
-	for (iI = 0; iI < GC.getNumUnitClassInfos(); iI++)
+	for (iI = 0; iI < GC.getNumUnitInfos(); iI++)
 	{
-		eLoopUnit = ((UnitTypes)(kCivilizationInfo.getCivilizationUnits(iI)));
+		eLoopUnit = (UnitTypes) iI;
 
 		if (eLoopUnit != NO_UNIT)
```
billw2012: This is impossible. NO_UNIT is -1.

billw2012:
To clarify, this change is intended to adjust the construction value of units based on what buildings they might be able to construct, which currently is not considered?

Sources/CvCityAI.cpp
Toffer90: True, my change made that if check redundant, I tried to change as little as possible code for it to be easier to understand the real changes.

PR #166, opened 2019-09-22
Tech tree rewrite that reduce general RAM usage by 70 MB.

Toffer90:
There is practically no persistent tech tree caching anymore, when you exit the tech screen all that it has cached is less than 1 MB.


billw2012:
We should take this now v40 is out right?

PR #202, opened 2019-10-25
Tech tree paging

billw2012:


billw2012:
@Toffer90 You can have a look at it now if you want, it is in final state, except for mouse scroll wheel (which I'm doing now), and there is a lock up when using it in advanced start that I didn't fix yet.

Toffer90:
K, I will probably look at it before the next weekend is over.

PR #210, opened 2019-10-31
feat(Language): german translations to some of the new pedia labels

dsmabell:
My first pull request. I keep it small, cause it's mainly a test, if I made everything right. 
Content: Some new translations for latitude and auto-build pedia labels and some minor corrections
- the history label should now display correctly
- German newline instead of paragraph to suppress hanging feeder (not used in German)


billw2012:
Good job! Even using the commit prefixes correctly 👍 
If you want to add more stuff before we accept it then just commit more to your fork and they will be added here automatically.

PR #213, opened 2019-11-02
Changed a bunch of tech names and pedia entries

GreatCoffers:


billw2012:
This looks good technically, not sure who wants to check the changes themselves.

billw2012:
Ignore the "Some checks were not successful" it looks like there is a bug in that system when pull request is from a fork.

raxo2222:
Looks like we should use spellchecker after this gets merged. By the way we already have space tourism tech.

KaTiON-PT:
>By the way we already have space tourism tech.

Correct, Commercial Spaceflight represents how private enterprises start to go to space and "commercialize" it (think SpaceX, VirginGalactic,Blue Origin).

KaTiON-PT:
By the way, @GreatCoffers I suggest changing "Direct Democracy" to "[Constitutionalism](https://en.wikipedia.org/wiki/Constitutionalism)".

GreatCoffers:
Constitutionalism would be in the late Renaissance around Constitution.  Direct Democracy is about the Greek version of democracy, where citizens get to vote directly rather than vote representatives into office. It should technically be called Republics to reflect the civic

KaTiON-PT:
Ah my bad, thought it was the modern democracy one.

alberts2:
You are only changing the English texts that invalidates the texts for the other Languages.

You should either change the translations as well or remove the invalid translations.

raxo2222:
Alberts, that happened countless times - for example all civic texts were changed, so now civic translations are invalid.
I bet majority of translations in pedia/quote/stratygy texts are invalid.

alberts2:
> Alberts, that happened countless times - for example all civic texts were changed, so now civic translations are invalid.

That's a problem that should be fixed!!!!

raxo2222:
> 
> 
> > Alberts, that happened countless times - for example all civic texts were changed, so now civic translations are invalid.
> 
> That's a problem that should be fixed!!!!

Well it is going to be tricky to fix 10 years of carelessness with translations.
I added updating/fixing translations point to text fixes project.

Toffer90:
It is common sense to remove incorrect translations, and its been pointed out in the forum countless of times,
I would think most C2C modders who modify text follows this rule.
The text change needs only to be significant enough to slightly the meaning for this rule to apply.

raxo2222:
Well then @KaTiON-PT should have to remove translations of ALL civics for example, as he overhauled their history/strategy entries
Similarly tech history/strategy/quotes would have to be removed.

I think no one except DH was taking care of translations.

GreatCoffers:
https://docs.google.com/document/d/1hQeWf8sAG5_lbmsMNreaa9SlSVhhXwo-MuKW6yeb1ss/edit?usp=sharing

Doc of my changes.

PR #226, opened 2019-11-10
fix(pedia): sun, brown bears entry no longer states they make bird buildings

dzfischer:
Sun and Brown bears had a pedia entry suggesting they construct bird buildings, not bear ones (their actual functionality was fine). Tiny change, mostly to see if I can figure out git kraken / how best to submit text fixes I find


raxo2222:
Looks good, it seems like DH slipped finger on this one lol.

PR #234, opened 2019-11-11
fix(Buildings): remove Tradition - Hunting as it isn't buildable

billw2012:


billw2012:
@Thunderbrd can you just ok this before I commit it? Apparently this thing was showing up but there is no way to make it. Apparently abandoned part way through development, now just confusing people.

Thunderbrd:
Are you sure it's not used in an event?

billw2012:
I could not any reference to it anywhere in xml, python or dll, I will check again though.

Thunderbrd:
The naming suggests it is supposed to be something a General or Great Hunter can make free to units OR that it comes from an event.  You pointed out that there are 2 versions and this one looks like a WIP that was never fully used.  If you've not been able to find any reference to it in any xml, python or dll file, then obviously, the other one is in use and this may just be one that was made and then forgotten so I'd be ok with removing it - it's on record and could be recovered.

PR #235, opened 2019-11-11
fix(wonder prereqs): mostly animal trainer+heritages and units, plus log roller

dzfischer:
This should be it for matching the alternate history animal trainer wonders and units with each other (plus the log roller because it annoyed me once), though the entire animal trainer module really needs a polishing pass or two; it's all over the place, both in coding practices and pedia text

### DIFF
--- a/Assets/Modules/Alt_Timelines/Megafauna_Units/Bear/Bear_CIV4BuildingInfos.xml
+++ b/Assets/Modules/Alt_Timelines/Megafauna_Units/Bear/Bear_CIV4BuildingInfos.xml
@@ -61,47 +61,47 @@
 					<bPrereqBuildingClass>1</bPrereqBuildingClass>
 				</PrereqOrBuildingClass>
 				<PrereqOrBuildingClass>
-					<BuildingClassType>BUILDINGCLASS_ASIANBEAR_MYTH_EFFECT</BuildingClassType>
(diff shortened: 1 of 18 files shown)


raxo2222:
I checked it, looks good enough, but things should require myth EFFECTS, not myths themselves.
Myth effects are created everywhere in empire, while myth is in one city only. 
Myth and stories effect makes those myth and stories pseudonational wonders.

dzfischer:
The animal trainers all required myths instead of myth effects with the exception of the bear trainer, so that should be the only one I switched from myth effect to straight myth, to align with the other trainers. Easy enough to switch back if you want?

raxo2222:
Well I'm replacing myth with myth effect requirements now, so no need for this.

raxo2222:
It appears only megafauna domestication buildings weren't using myth effects lol.

dzfischer:
Yeah, that was the feeling I got. It kinda made sense in some regards, as in 'Only where the myth itself was born, could a great trainer rise to the occasion!' but, either way works fine

raxo2222:
Megafauna domestication is not too far from Tribalism, so you will have only couple of cities anyway unless you go all out of war on barbarian cities and neighbour civs.
Also myth effects appear instantly anywhere, so there is no reason to force those buildings to single city ;)

PR #257, opened 2019-11-21
fix(audio): mp tech quotes now either vanilla mp quotes or just 'ding'

dzfischer:
As Thunderbird suggested, multiplayer quote sfx are now either the "You have discovered X" from vanilla quotes where appropriate, or just the 'ding' sfx in place of potentially long, drawn out spoken quotes.

I initially wanted to rename all the tech .mp3 files to have a prefix matching the era the tech is in, for ease of editing them later (and also just for consistency's sake since they're all over the place...) but that didn't quite work out. So, guess I'm doing that manually now in a separate project as I'm passing through recording quotes (will make a different PR as I finish prehistoric era).

### DIFF
--- a/Assets/Modules/Alt_Timelines/Atompunk/Atompunk_CIV4TechInfos.xml
+++ b/Assets/Modules/Alt_Timelines/Atompunk/Atompunk_CIV4TechInfos.xml
@@ -10,7 +10,7 @@
 			<Advisor>ADVISOR_CULTURE</Advisor>
 			<Quote>TXT_KEY_CULTURE_ATOMPUNK_QUOTE</Quote>
 			<Sound>AS2D_TECH_ATOMPUNK</Sound>
-			<SoundMP>AS2D_TECH_ATOMPUNK</SoundMP>
+			<SoundMP>AS2D_TECH_GENERIC</SoundMP>
 			<Button>Art/Interface/Buttons/TechTree/retrospace.dds</Button>
 			<!-- Main properties -->
 			<iCost>13181</iCost>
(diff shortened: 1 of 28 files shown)


billw2012:
Thanks although there is no such commit tag as `tweak` the valid values are `fix`, `feat`, `balance` and `perf`.

dzfischer:
Heh, what... should it be then? Fix, maybe? I also just realized it looks like that's how the changelogs on cfc are generated, so should I not be using that  commit style for minor changes like this?

billw2012:
I would say `fix`, you can actually rename it if you want to. Yes the tags are used to generate the changelog formatting. All changes go in the log, but only those correctly formatted with a supported tag will be highlighted in the matching sections. The `(<blah>)` bit is just an extra call out for the general area the changes relate to, and as such should be a capitalized noun ideally. I would say `fix(Audio):` would be the correct prefix. 
There is no change to small to call out if it changes the player experience. The only commits that should *not* have these prefixes are ones that won't change the player experience at all (refactoring, tools additions etc.)

billw2012:
The Pull Request Labeler fail can be ignored in case its not obvious! It breaks for PRs from forks.

billw2012:
The conversion script should go into our `Tools\XMLTools` directory not in the Assets directory please!

billw2012:
This is fine with me, @Thunderbrd you can merge it when you are happy?

Thunderbrd:
I probably won't have time to review this soon and it seems to be a fairly minor adjustment that is designed to improve the streamlining of the mod design.  I'm all for it based on the assumption it's been done right and it's a lot to review just to make sure of that.  So since my stuff is out of date and I'm not in a position to update it right at the moment, go for it - if you approve, I approve.

PR #268, opened 2019-11-23
Deactivated nonsensical event

raxo2222:
Now it shouldn't trigger anymore

### DIFF
--- a/Assets/XML/Events/CIV4EventTriggerInfos.xml
+++ b/Assets/XML/Events/CIV4EventTriggerInfos.xml
@@ -9369,8 +9369,8 @@
 					<Text>TXT_KEY_EVENT_CAKE_FLOUR</Text>
 				</TriggerText>
 			</TriggerTexts>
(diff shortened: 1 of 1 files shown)


Toffer90:
Proper way to deactivate it is to set iPercentGamesActive = -1
0 actually means it will be active in 1% of all games as 99 is active in 100 percent of games.
The code could be changed so the xml value make more sense, I agree, just how it is currently.

iWieght 0 is correct though, as -1 here means that the event will force trigger whenever its requirements are fulfilled. I don't think the iWeight value matters if iPercentGamesActive = -1 though, there may be quirks in the code so setting it to 0 is sensible.

raxo2222:
In my test modmod I disabled annoying events by setting iPercentGamesActive and iWeight to 0.
It worked for me, so I used those values ;)

Toffer90:
Ah sorry, I misread the codeline 0 means it will always be deactivated too.

PR #271, opened 2019-11-24
Removed unused code.

MattOttawa:


billw2012:
Looks okay, whats wrong with the `search strings.txt` file? I have never seen it before, but I assume someone wanted it.
You checked that the changes to the Python interfaces didn't require any updates to the Python code?

Toffer90:
This is MattCA right?
Thanks for getting rid of the DCM and AND concept stuff for the pedia, that is something I removed all usage of in PPIO; but PPIO was a python only modmod so I didn't touch the dll setup for it; I had actually forgotten there was unused leftovers in the dll for that stuff.

MattOttawa:
Search Strings.txt has about a dozen examples of tags a modder may want to search for on the dll.
Because there is a million different things to search for I don't see those examples as helpful.
I'd call it clutter. 
File was added svn rev 5, along with all ROM's source files i think.

I took a second to search the python, although I'm aware Toffer took care of this stuff in python.
Loose ends remaining would be the 2 empty xml files in BasicInfos(?) for the pedia concepts.

PR #273, opened 2019-11-24
fix(audio): rerecorded poor quality prehistoric era singleplayer tech quotes

dzfischer:
Accidentally started this branch off of the mp tech quote pr earlier, so there commit history is a little awkward with 3 initial nonrelevant commits and 3 final reverts of the initial commits, but should check out.


Toffer90:
Thumb up.

PR #279, opened 2019-11-27
code cleanup

MattOttawa:
CvGame: Added const to methods.
CvGame: Moved function variable declarations. (let me know if that's too annoying)
CvGame: Reduced a loop's size from MAX_PLAYERS (51) to MAX_PC_PLAYERS (40) and removed if(!isNPC()).
CvGame: Removed if(iter != NO_PLAYER) from a loop that starts with 0.
CvUnit: Changed GC.getDefineINT("EVENT_MESSAGE_TIME") calls to GC.getEVENT_MESSAGE_TIME()
CvUnit: Removed a pointless function variable named bBitter from about 5 DCM airBomb methods.

- I don't think bBitter is an upcoming TB thing, but I can change this pull request if I'm wrong.

### DIFF
--- a/Sources/CvGame.cpp
+++ b/Sources/CvGame.cpp
@@ -15,10 +15,6 @@
 /************************************************************************************************/
 /* BETTER_BTS_AI_MOD                       END                                                  */
 /************************************************************************************************/
-// BUG - start
-// RevolutionDCM - BugMod included in CvGlobals.h
-//#include "BugMod.h"
-// BUG - end
 
 //	Koshling - save game compatibility between (most) builds
 //	UI flag values in game serialization.  These are bitwise combinable
@@ -124,7 +120,6 @@ CvGame::~CvGame()
 
 void CvGame::init(HandicapTypes eHandicap)
 {
-	bool bValid;
 	int iStartTurn;
 	int iEstimateEndTurn;
 	int iI;
@@ -323,26 +318,15 @@ void CvGame::init(HandicapTypes eHandicap)
 	// need to be "enabled" so starships can be built, etc. etc.
 	//Only exception is diplo victory, but whatever.
 
-	
 		if (getEstimateEndTurn() > getGameTurn())
 		{
-			bValid = false;
-
 			for (iI = 0; iI < GC.getNumVictoryInfos(); iI++)
 			{
-				if (isVictoryValid((VictoryTypes)iI))
+				if (isVictoryValid((VictoryTypes)iI) && (GC.getVictoryInfo((VictoryTypes)iI).isEndScore() || GC.getVictoryInfo((VictoryTypes)iI).isTotalVictory()))
 				{
-					if ((GC.getVictoryInfo((VictoryTypes)iI).isEndScore()) || (GC.getVictoryInfo((VictoryTypes)iI).isTotalVictory()))
-					{
-						bValid = true;
-						break;
-					}
+					setMaxTurns(getEstimateEndTurn() - getGameTurn());
+					break;
 				}
-			}
-
-			if (bValid)
-			{
-				setMaxTurns(getEstimateEndTurn() - getGameTurn());
 			}
 		}
 	}
@@ -355,24 +339,17 @@ void CvGame::init(HandicapTypes eHandicap)
 	// If Total Victory is enabled, TURN ON ALL OTHERS AS WELL
 	// This allows starship construction, gets the AI thinking correctly, etc.
 	// The TV condition overrides the others but requires they be "on"
-	bValid = false;
 	for (iI = 0; iI < GC.getNumVictoryInfos(); iI++)
 	{
-		if ((GC.getVictoryInfo((VictoryTypes)iI).isTotalVictory()) && (isVictoryValid((VictoryTypes)iI)))
+		if (GC.getVictoryInfo((VictoryTypes)iI).isTotalVictory() && isVictoryValid((VictoryTypes)iI))
 		{
-			bValid = true;
+			for (iI = 0; iI < GC.getNumVictoryInfos(); iI++)
+			{
+				setVictoryValid((VictoryTypes)iI, true);
+			}
 			break;
 		}
 	}
-
-	if (bValid)
-	{
-		for (iI = 0; iI < GC.getNumVictoryInfos(); iI++)
-		{
-			setVictoryValid((VictoryTypes)iI, true);
-		}
-	}
-
 //Sevo End VCM
 #else
   }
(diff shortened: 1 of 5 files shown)


Sources/CvGame.cpp
```diff
@@ -4339,7 +4309,7 @@ int CvGame::getTurnSlice() const
 
 int CvGame::getMinutesPlayed() const
 {
-	return (getTurnSlice() / gDLL->getTurnsPerMinute());
```
billw2012: I don't think this is an improvement. Abstracting the value access is more flexible, and has no overhead (the compiler will inline it). In this case it probably doesn't matter, but I would not suggest this as a general improvement to apply everywhere.

Sources/CvGame.cpp
```diff
@@ -6445,17 +6397,17 @@ int CvGame::countNumReligionTechsDiscovered()
...
 	FAssertMsg(eIndex < GC.getNumCorporationInfos(), "eIndex is expected to be within maximum bounds (invalid Index)");
 	return m_paiCorporationGameTurnFounded[eIndex];
 }
 
 
-bool CvGame::isCorporationFounded(CorporationTypes eIndex)
+bool CvGame::isCorporationFounded(CorporationTypes eIndex) const
 {
-	return (getCorporationGameTurnFounded(eIndex) != -1);
+	return (m_paiCorporationGameTurnFounded[eIndex] != -1);
```
billw2012: Not an improvement as per my explanation above.

Sources/CvUnit.cpp
```diff
@@ -31544,99 +31533,96 @@ bool CvUnit::airBomb3(int iX, int iY)
 			}
 		}
 	}
-	if (bBitter)
```
billw2012: Weird that static analysis didn't pick this up!

Sources/CvGame.cpp
```diff
@@ -3799,26 +3768,23 @@ int CvGame::countTotalNukeUnits()
...
 {
 	int iCount = 0;
 
-	for (int iI = 0; iI < MAX_TEAMS; iI++)
+	for (int iI = 0; iI < MAX_PC_TEAMS; iI++)
 	{
-		if (GET_TEAM((TeamTypes)iI).isEverAlive())
+		if (GET_TEAM((TeamTypes)iI).isEverAlive() && GET_TEAM((TeamTypes)iI).isHasTech(eTech))
 		{
-			if (GET_TEAM((TeamTypes)iI).isHasTech(eTech) && !GET_TEAM((TeamTypes)iI).isNPC())
```
billw2012: What is this change about? I can guess the ordering improves perf, but why `!isNPC` to `isEverAlive`?

Sources/CvGame.cpp
```diff
@@ -13257,25 +13152,21 @@ void CvGame::findBays()
 void CvGame::findForests()
 {
 	CvPlot* pLoopPlot;
-	int iForestCount;
+	FeatureTypes eForest = (FeatureTypes)GC.getInfoTypeForString("FEATURE_FOREST");
```
billw2012: This should be `const`.

Sources/CvGame.cpp
Toffer90: isEverAlive was the next if test inside that if,, he concatenated two ifs when he removed the isNPC check.
Since he change the for loop to stop before reaching the NPC teams there is no need to ask if the looped player is an NPC as the answer is already known.

MattOttawa: adding && and removing !isNPC() in the same place makes things look funky, but ya what toffer said. There was a short discussion on the forum a while back about some player and team loops having wrong size and stuff. That's a good example.

Sources/CvGame.cpp
MattOttawa: done

MattOttawa: how reliable is the compiler if i add const in the wrong spot?

billw2012: Not certain what do you mean? It will 100% not compile if you try to modify a const object if that is what you mean. However the declaration of const is something you as a coder do to express your intent in code. Even if the compiler didn't care it would still be useful as it gives more information about the variables intention (or functions). 
However in the case of const the compiler cares a **lot**. It will strictly enforce it (unless you are subverting it with `mutable` which should be *very* rare), which helps a lot with enforcing logical consistency. It will also help the optimizer when it comes to deciding what optimization it can perform (you can see an example here: https://www.youtube.com/watch?v=zBkNBP00wJE&t=1635s).

Sources/CvGame.cpp
```diff
@@ -3930,19 +3835,11 @@ bool CvGame::canTrainNukes() const
 		CvPlayer& kPlayer = GET_PLAYER((PlayerTypes)iI);
 		if (kPlayer.isAlive())
 		{
-			for (int iJ = 0; iJ < GC.getNumUnitClassInfos(); iJ++)
+			for (int iJ = 0; iJ < GC.getNumUnitInfos(); iJ++)
 			{
-				UnitTypes eUnit = (UnitTypes)GC.getCivilizationInfo(kPlayer.getCivilizationType()).getCivilizationUnits((UnitClassTypes)iJ);
```
billw2012: Is this check not necessary then? I have seen it in *many* places, but never looked into what it really means.

Sources/CvGame.cpp
billw2012: Oh I see, I didn't know what MAX_PC_TEAMS was. Is PC meant to be Player controlled? Would be clearer if the name didn't have an ambiguous acronym in it.

Sources/CvGame.cpp
Toffer90: C2C doesn't have civilization specific units, so regardless of what civ it is checking - a given unitclass type for - it will always return the same unit type for it.

Civ specific units depend upon the unitclass object; which we intend to remove eventually.

Sources/CvGame.cpp
Toffer90: Playable Civilizaation/Character
As opposed to NPC (Non Playable Character)

billw2012:
@MattOttawa It looks good then, are you ready for it to go in or do you have more stuff you want to add?

Sources/CvGame.cpp
Thunderbrd: PC vs NPC is a pretty standard gaming naming convention, however it often usually considers NPC to be any non-human and we have actually considered the Non-Human players in the PC side as long as they are not a non-standard 'PLAYER' in the game.  NPCs are Barbs, Animals, etc...

MattOttawa:
If you're good with everything go ahead and merge.

PR #293, opened 2019-12-01
Removed GAMEOPTION_MOUNTAINS

MattOttawa:
No intentional in-game changes.


Thunderbrd:
This has been discussed on the forums many times and it has always come up with needing to keep the option based on community consensus.  At least available in the background.  Also, removing an option will invalidate all games saved - not an outright incompatibility but it does not adapt well as it will throw off the boolean delineation of options on all options set after the one that's been removed.  Therefore, removing an option should only be done when we are disrupting save compatibility in the first place.

That said, Dancing Hoskuld has always strongly advocated that the option exist, and the fact that it is a major adjustment from original vanilla still makes it something some players would prefer to be able to turn off.

MattOttawa:
np bro.

PR #294, opened 2019-12-02
Removed LAST_PLAYER

MattOttawa:


raxo2222:
When you are at it can you make barbarian cities not spawn on space terrains?
That is it seems if hill is nearby, then barbarian cities will spawn near it.

PR #296, opened 2019-12-03
Option: disable automated hunting units return to city for upgrade

billw2012:


billw2012:
Okay well I will merge this if no one can review it!

PR #309, opened 2019-12-16
fix(tech prereq): Hunting now requires Tracking

dzfischer:
Specifically, "Hunting Tactics" requires "Hunting". Also fixed a teensie typo for another tech quote.


dzfischer:
Whoops, fixed that. Didn't realize that's what controlled the arrows

raxo2222:
Now its good.

PR #315, opened 2019-12-21
pr297-merge revised.

MattOttawa:


Sources/CvPlayerAI.cpp
```diff
@@ -31373,61 +31274,73 @@ CvCity* CvPlayerAI::getTeamInquisitionRevoltCity(CvUnit *pUnit, bool bNoUnit, in
 CvCity* CvPlayerAI::getReligiousVictoryTarget(CvUnit *pUnit, bool bNoUnit)
 {
 	FAssert(pUnit != NULL);
-
-	if(!hasInquisitionTarget() || (!isPushReligiousVictory() && !isConsiderReligiousVictory()))
```
billw2012: You undid a bunch of changes here, why?

Sources/CvSelectionGroupAI.cpp
```diff
@@ -757,10 +709,13 @@ CvUnit* CvSelectionGroupAI::AI_getBestGroupAttacker(const CvPlot* pPlot, bool bP
...
 		// if the battle occurs.  This will return a better odds estimate than we
 		//	had previously, but should not change the unit choice
-		iBestOdds = pBestUnit->AI_attackOdds(pPlot, bPotentialEnemy, ppDefender);
+		iUnitOdds = pBestUnit->AI_attackOdds(pPlot, bPotentialEnemy, ppDefender);
 	}
-	
-	iUnitOdds = iBestOdds;
+	else
+	{
+		iUnitOdds = iBestValue;
```
billw2012: Odds aren't the same as the value before the changes, are you sure this is correct?

MattOttawa: This bit is rather odd actually.  iUnitOdds doesn't seem to be used. If pBestUnit->AI_attackOdds(pPlot, bPotentialEnemy, ppDefender) does nothing exceptreturn something, the whole part can be removed.

Sources/CvSelectionGroupAI.cpp
```diff
@@ -669,67 +646,43 @@ CvUnit* CvSelectionGroupAI::AI_getBestGroupAttacker(const CvPlot* pPlot, bool bP
...
 			{
-				if (bForce || pLoopUnit->canMove())
+				if ((!pLoopUnit->AI_getHasAttacked() || bSuprise) && (bForce || pLoopUnit->canMove()))
 				{
 					CvUnit* pBestDefender = NULL;
 					if (bForce || pLoopUnit->canMoveInto(pPlot, moveCheckFlags, &pBestDefender))
 					{
 						PROFILE("AI_getBestGroupAttacker.RegularAttackOdds");
 
-						const int iOdds = pBestDefender?
```
billw2012: I'm talking about here: `iOdds` is the value returned by `AI_attack*` functions, `iValue` is the same but further modified below. That makes `iBestOdds` and `iBestValue` different, so you can't just remove iBestOdds. Did I miss something here?

billw2012:
Why did you close it? There was only one more question remaining to answer!

MattOttawa:
9 conflicts after commit to revert that.  I'll just get rid of more files.

raxo2222:
What you tried to do?
It was meant to remove spawning of barbarian cities outside Earth mapcategory tiles - currently hills allow barbarians to spawn in space

MattOttawa:
I'll get that in the mod but it's turned off... actually thinking now, I'll make an xml global define...

raxo2222:
I mean 42 files were changed a lot

MattOttawa:
haha ya, this pr got way too big, was a mistake.

billw2012:
It was SO close! Could have accepted it and then fixed it myself.
Also you don't need to recreate your fork every time! What is that about?
Also PLEASE join discord, we can talk about this then I would have said to just wait and I will merge it myself instead of you having to redo it.

MattOttawa:
haha sorry man, I don't know how to use github.  I'm pretty sure i still have it and I will fix it all.

billw2012:
Wait, you are going to redo the entire thing from scratch? Or you saved it some how? Please tell me you saved it somehow!

PR #317, opened 2019-12-21
Small clean up in the usable mountain GM code.

Toffer90:
Changes make sense to me, though I would like a second opinion.
Was a bug in the AI code that add value of the mountaneering tech specifically by counting up all mountains inside a players culture, it only did this if usable mountain option was turned off, which is moot as the tech could not even be invented if that option was off.
Optimized away a couple loop through all techs in AI valuation of promotions and unit combat types related to mountaineering.
Streamlined what the option does when turned off:
1. Tech is now available even when the option is off. It unlocks improvements, buildings and promos that I see no reason to optionalize in any way.
2. Inventing the tech won't allow all owned units to pass mountains like how it is with the option on. 
3. Land units must have relevant promos or unit combat types to pass mountains, no exceptions.
4. Founding cities on mountains is impossible with the option off. (I believe the city graphic looks glitchy/bad on peaks)

P.S. Units have for several years been capable of traversing and routing peaks when usable mountains was turned off, I never played with it on while it was a visible option as I didn't like certain aspects of it, and I still got promotions that allowed units to move through peaks.

### DIFF
--- a/Assets/XML/Technologies/CIV4TechInfos.xml
+++ b/Assets/XML/Technologies/CIV4TechInfos.xml
@@ -12156,6 +12156,9 @@ Current techs which are not in this file:-
 			<OrPreReqs>
 				<PrereqTech>TECH_ASTROLABE</PrereqTech>
 			</OrPreReqs>
+			<AndPreReqs>
+				<PrereqTech>TECH_MOUNTAINEERING</PrereqTech>
+			</AndPreReqs>
 			<Quote>TXT_KEY_TECH_ASTRONOMY_QUOTE</Quote>
 			<Sound>AS2D_TECH_ASTRONOMY</Sound>
 			<SoundMP>AS2D_TECH_MP_ASTRONOMY</SoundMP>
--- a/Sources/CvGame.cpp
+++ b/Sources/CvGame.cpp
@@ -13785,10 +13785,6 @@ int CvGame::getAverageCorporationInfluence(CvCity* pCity, CorporationTypes eCorp
 
 bool CvGame::canEverResearch(TechTypes eTech) const
 {
-	if (!isOption(GAMEOPTION_MOUNTAINS) && GC.getTechInfo(eTech).isCanPassPeaks())
-	{
-		return false;
-	}
 	if (isOption(GAMEOPTION_NO_FUTURE))
 	{
 		if (GC.getTechInfo(eTech).getEra() > GC.getInfoTypeForString("ERA_MODERN"))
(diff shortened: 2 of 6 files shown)


Toffer90:
I started work on these modifiaction 90 minutes ago, and I've been awake 24 hours now, so there may be some bad logic to some of my changes, haven't tested it in game either.
I would love if raxo or someone else would be interested in testing out the mountain stuff both with and without the option enabled on a fast gamespeed using this branch; I don't have much time for testing stuff in the near future.

raxo2222:
Since Mountaineering isn't dead end tech anymore, it could be required by one of following techs:
Astronomy (Renaissance)
Geology (Renaissance)
Tourism (Atomic)
Modern Sports (Atomic)

Toffer90:
It is a dead end tech still, doesn't need to not be a dead en tech. If something make sense then sure add it as a prereq to some tech, could simply be required by only one tech later in the tech tree without an arrow in between; whatever makes sense.

raxo2222:
I meant to write optional, not dead end in that comment :p

Only techs, that are dead end are punks, religious techs, waterproof concrete/crystal glassware (those two require buildings), and mountaineering as this tech could be disabled with option.

Toffer90:
I personally think there should be more dead ends in the tree, and the dead could even be more than just one tech, may be a whole dead branch in the tree if it makes sense. Could add more layers of strategy if the dead branches are designed to be a short sighted juicy goals when the situation for it is right, like having the right resources or cultures and such. Could be a faster way to an alternative unit that is stronger than the next strong unit in the long sighted living part of the tree, going for it will delay your progression towards the unit that is stronger than that again in the living tree. Stuff like that.

raxo2222:
This game option was set to be hidden and enabled by default as requested some time ago by Thunderbrd

Astronomy will be tech, that requires mountaineering.

raxo2222:
Looks like settlers can't found cities on mountains even when option is ON.

raxo2222:
There is bug: You can't build mountain mine on peak.

Toffer90:
> There is bug: You can't build mountain mine on peak.

K, thanks for checking that, I'll look into why that may be the case.
Was that with, or without, the option enabled, or in both cases?

raxo2222:
This happens if option is enabled, didn't check it what happens if option is disabled.

Toffer90:
> There is bug: You can't build mountain mine on peak.

I can't reproduce it with the save you provided, I could build mountain mines just fine there on this branch.

raxo2222:
I guess it was one off glith then

PR #327, opened 2020-01-01
Animal AI Refinement

Toffer90:


billw2012:
Please provide some description of the intent of the changes!

Thunderbrd:
Removing all the AI regarding if the unit is not an animal is removing all the hunter leading back a captured animal coding if I'm not mistaken but I'm not 100% sure on that.

Toffer90:
> Removing all the AI regarding if the unit is not an animal is removing all the hunter leading back a captured animal coding if I'm not mistaken but I'm not 100% sure on that.

I thought it was legacy code due to some subdued animals having UNITAI_ANIMAL defined in their xml instead of the newer UNITAI_SUBDUED_ANIMAL. Considering there was a FAssertMsg(false, "Subdued animal still has animal AI"); an assert that always triggers there made me think it was.
I've never seen this assert trigger so I believe it is quite uncommon for non animal NPC teams to have any units who use the UNITAI_ANIMAL AI routine.

Thunderbrd:
hmm... you may be right.  I haven't looked too deeply into that.  Alberts2 may know more from direct work on that.

Toffer90:
> Please provide some description of the intent of the changes!

Right, it was mostly a tweak to make the animalAttackProb value from the handicap xml tag behave more like one would expect and to fine tune animal aggressiveness a bit.

I also re-added the commented out flee routine to wild animals AI_safety() if not playing on reckless animal options and the animal felt it was exposed to danger [exposedToDanger()] after it had figured out that it didn't want to attack something.

Toffer90:
I would also like to expand the aggression value range from 0-10 to 0-100 making it more a percentage where each unit level above the first adds 4 point of aggression (4%) instead of the current 10%.
I feel the 10 step aggression level is too restrictive to vary animal aggression between animal types and through other modifiers like promotions/levels for it to provide meaningful results.

Though I delayed doing that in this PR to restrict the amount of changes in this PR.

Thunderbrd:
I don't mind if you want to make it more granular.  My only question on all this was if the first part can so safely be removed.  But you can try it and we'll see if we get some odd behavior that accounts for it.

Sources/CvUnitAI.cpp
```diff
@@ -1866,84 +1866,51 @@ void CvUnitAI::AI_animalMove()
...
+		if ((iOddsThreshold < 1 || iAttackProb > 99 || GC.getGameINLINE().getSorenRandNum(100, "Animal Attack") < iAttackProb)
+			&& AI_attackTargets(2, iOddsThreshold, 0, false, true))
 		{
-			OutputDebugString("Chosen to construct\n");
 			return;
 		}
 
-		if (AI_guardCity())
+		//Then attack elsewhere, recklessness based on animal aggression.
+		int iAggression = 666;
```
billw2012: What is the meaning of this magic number? I note you compare against it below, I'm guessing this is meant to be used as a flag to indicate that iAggression wasn't changed, but it doesn't make the intent clear at all. 
Shouldn't it just be 
`if (bReckless || GC.getGameINLINE().getSorenRandNum(10, "Animal Attack") < iAggression)` 
on line 1891 and 
`if (!bReckless && exposedToDanger(plot(), 60, true) && AI_safety())`
on line 1900?
And if you want to improve it further pull out the iAggression change to a separate function so it would be 
`if (bReckless || GC.getGameINLINE().getSorenRandNum(10, "Animal Attack") < getAggression())`

Toffer90:
> I don't mind if you want to make it more granular. My only question on all this was if the first part can so safely be removed. But you can try it and we'll see if we get some odd behavior that accounts for it.

It can safely be removed, only the wild animal units use the UNITAI_ANIMAL and no non-animal teams can ever get these wild animal units, and there is no code that sets that AI for units that doesn't have it defined in their xml.

The first part I removed would trigger if you in the worldbuilder added wild animal unit to one of the non-NPC AI teams, or changed the unitAI of one of their regular units to it in WB,, but I see no reason why UNITAI_ANIMAL should behave differently for non-NPC teams than it does for NPC teams. If there needs to be a difference in the unit AI routine based on what type of team owned the unit then it should be a minor difference, a tweak inside the routine, not a complete rewrite of the entire routine.

Sources/CvUnitAI.cpp
Toffer90: The number indicates that the reckless animal option is active and that the animal should always attack.

Toffer90: I guess it could be optimized so that the sorenRandNuimber function isn't called if bReckless is True.

Toffer90: Hmm, trying to add the new agression function and get this on compile 
error C2039: 'getMyAggression' : is not a member of 'CvUnitAI'
Do I need to add it to the .h file, or modify something elsewhere?

I added the function right above the AI_animalMove function.
```cpp
int CvUnitAI::getMyAggression()
{
	int iAggression = m_pUnitInfo->getAggression() + getLevel() - 1;
	iAggression *= maxHitPoints() - getDamage();
	iAggression /= 100;
	iAggression *= iAttackProb;
	return iAggression / 100;
}

// Protected Functions...

void CvUnitAI::AI_animalMove()
```

Toffer90: lol, I'm missing variables here, forget what I said. lol

Toffer90: Ok, figured out how to add the .h part and now it compiles.

Sources/CvUnitAI.h
```diff
@@ -528,6 +528,7 @@ class CvUnitAI : public CvUnit
 	bool isWaitingOnUnitAIAny();
 	void setWaitingOnUnitAIAny();
 	bool AI_isNegativePropertyUnit();
+	int getMyAggression(int iAttackProb);
```
billw2012: Last thing: it should be a const function as it does not modify the object members.
`int getMyAggression(int iAttackProb) const;`
also add const in the cpp in the same way.

Toffer90: Now I get 
error C2511: 'void CvUnitAI::AI_animalMove(void) const' : overloaded member function not found in 'CvUnitAI'

What did I do wrong?

Toffer90: ah lol, I'm tired, added const to wrong function in cpp.

billw2012:
Well I will merge it then

PR #336, opened 2020-01-05
[WIP] Removing INLINE from function names.

MattOttawa:
- work in progress.

I'm just changing names. Everything is still declared inline.

### DIFF
--- a/Sources/CvMap.cpp
+++ b/Sources/CvMap.cpp
@@ -109,11 +109,11 @@ void CvMap::init(CvMapInitData* pInitInfo/*=NULL*/)
 	//--------------------------------
 	// Init other game data
 	gDLL->logMemState("CvMap before init plots");
-	m_pMapPlots = new CvPlot[numPlotsINLINE()];
-	for (int iX = 0; iX < getGridWidthINLINE(); iX++)
+	m_pMapPlots = new CvPlot[numPlots()];
+	for (int iX = 0; iX < getGridWidth(); iX++)
 	{
 		gDLL->callUpdater();
-		for (int iY = 0; iY < getGridHeightINLINE(); iY++)
+		for (int iY = 0; iY < getGridHeight(); iY++)
 		{
 			plotSorenINLINE(iX, iY)->init(iX, iY);
 		}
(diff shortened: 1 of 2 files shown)


billw2012:
Yeah nice, the INLINE versions are pointless, DLLExport functions [are inlinable](https://devblogs.microsoft.com/oldnewthing/?p=2123) so I don't know why someone wrote special INLINE versions anyway.

MattOttawa:
So I can get rid of the "NonInl" functions I added right?

billw2012:
Yep, only one function is necessary!

billw2012:
Do you not build before committing? In debug it should only take < 30s to do iterative build usually.

MattOttawa:
not with these draft ones. 
What's iterative build?
Here's from ust now.  It failed early.

1>--- Summary -----------------------------------------------------
1>                                 /----- Cache -----\
1>Build:          Seen    Built   Hit     Miss    Store   CPU
1> - Directory  : 2       2       -       -       -       0.063s
1> - Exec       : 1       1       -       -       -       10.036s
1> - File       : 1957    73      -       -       -       0.081s
1> - Object     : 63      9       -       -       -       1h:17m 58.185s
1> - Alias      : 1       0       -       -       -       0.000s
1> - Unity      : 1       1       -       -       -       4.783s
1> - Compiler   : 1       0       -       -       -       0.000s
1> - DLL        : 1       0       -       -       -       0.000s
1> - ObjectList : 1       0       -       -       -       0.000s
1>Cache:
1> - Hits       : 0 (0.0 %)
1> - Misses     : 0
1> - Stores     : 0
1>Time:
1> - Real       : 27m 30.273s
1> - Local CPU  : 1h:18m 13.148s (2.8:1)
1> - Remote CPU : 0.000s (0.0:1)
1>-----------------------------------------------------------------
1>FBuild : error : BUILD FAILED: Release-build
1>Time: 27m 34.089s
1>...Failed

raxo2222:
Use compile tools

billw2012:
What? If you are a coder just build in VS 2019, why would you use the batch file?! If you just want to test the compile then use Debug config.
This is a full rebuild in Debug:
image

Thunderbrd:
Question about that - with the batch build, it will automatically place the dll in the assets - will it do that automatically when building in the VS environment?

billw2012:
In both cases it will.

PR #338, opened 2020-01-06
Removed CvPythonModulesInfo.

MattOttawa:
- It's looking for xml files in a folder named "Python Config".
 - Is not used in the dll or python.


billw2012:
Is this not used by modmods?

raxo2222:
Not used by them - all active modmods are pure XML.

Issue #275, opened 2019-11-26
`resetUnitSMValues` is *only* called once during a session, not after every game is loaded.


billw2012:
`CvGame.cpp` line 2808.

billw2012:
Fixed in #339.

PR #339, opened 2020-01-06
SM cached values are now recalculated on load instead of once only

billw2012:
This attempts to address #275. 
Check out the comment in the code below for the motivation.


Sources/CvGame.cpp
```diff
@@ -2673,24 +2673,6 @@ void CvGame::update()
 		}
 	}
 
-	if (isOption(GAMEOPTION_SIZE_MATTERS))
```
billw2012: Not sure what this is doing, however the static bool below ensures it will only ever run once in a session (i.e. you need to restart the game to run it again). Therefore it could either be entirely redundant, OR bugged.

PR #340, opened 2020-01-06
[WIP] Just a quick guess about main menu era names.

MattOttawa:
I'm not 100% sure what the story is about the first era name but it's stuck a certain way cause of the main menu or something.
I think people don't mind it the way it is, so I'm just curious.

### DIFF
--- a/Assets/Python/DataStorage/ScreenResolution.py
+++ b/Assets/Python/DataStorage/ScreenResolution.py
@@ -12,7 +12,11 @@ def init(dir):
 	print "ScreenResolution.init\nSet custom resolution from CivilizationIV.ini if found."
 	import ConfigParser
 	Config = ConfigParser.ConfigParser()
-	Config.read(dir + "\CivilizationIV.ini")
+	path = dir + "\CivilizationIV.ini"
+	import IniMgr
+	IniMgr.readC2CIni(Config, path)
+
+	Config.read(path)
 	#Config.sections()
 	X0 = Config.get("DEBUG", "ScreenWidth")
 	Y0 = Config.get("DEBUG", "ScreenHeight")
--- a/Assets/Python/Matt/IniMgr.py
+++ b/Assets/Python/Matt/IniMgr.py
@@ -1,0 +1,7 @@
+from CvPythonExtensions import *
+
+def readC2CIni(Config, path):
+	Config.set("GAME", "Era", "ERA_INDUSTRIAL")
+	file = open(path, 'wb')
+	Config.write(file)
+	file.close()
--- a/Sources/CvDefines.h
+++ b/Sources/CvDefines.h
@@ -183,6 +183,8 @@
 
 #define ANIMATION_DEFAULT											(1)			// Default idle animation
 
+#define FIRST_ERA						((EraTypes)0)
+
 // python module names
 #define PYDebugToolModule			"CvDebugInterface"
 #define PYScreensModule				"CvScreensInterface"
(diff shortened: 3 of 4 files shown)


MattOttawa:
The type name in the xml I mean.

Thunderbrd:
The EXE fixed Ancient as the default starting era, rather than the enumerated first entry on the list.  If I recall that's what it was about.

MattOttawa:
Ya, that's right. This might be a stupid question but did u guys try switching the global define? It is set to ERA_ANCIENT atm.

MattOttawa:
the Standard Era one.

Toffer90:
The global define was set as ERA_PREHISTORIC before we renamed the prehistoric era to ERA_ANCIENT to fix the minor "bug" related to how last selected options are stored in the ini file.
I think the "bug" only happened after playing vanilla or another mod, then when you start C2C the ancient era would be the default era for new games.
It happened because starting a non-C2C game would set your last selected era in CivilizationIV.ini:
```ini
; Era options are ERA_ANCIENT/ERA_CLASSICAL/ERA_MEDIEVAL/ERA_RENAISSANCE/ERA_INDUSTRIAL/ERA_MODERN
Era = ERA_ANCIENT```
The exe sets the initial values in the ini file to ERA_ANCIENT after a clean install of BtS. We could make the C2C launching process reset the ini variable to whatever we want though.

**Comment by MattOttawa** [comment]
Interesting idea. I replaced 2 spot - one of them the data was originally coming from pSource.
I've got no idea where pSource data comes from, so could be from exe. Maybe...
I'll probly compile this and check at some point.

**Comment by Toffer90** [comment]
I tried adding this early in the python initialization:
```py
	import ConfigParser
	path = dir + "\CivilizationIV.ini"
	Config = ConfigParser.ConfigParser()
	Config.read(path)
	Config.set("GAME", "Era", "ERA_INDUSTRIAL")
	file = open(path, 'wb')
	Config.write(file)
```
And surely enough I forced industrial era to be the default selected era when choosing custom game after launching C2C.
If we wanted the game to remember the last selected era we could add **C2C_**ERA_X to all our era names and the python code would then only change the ini upon launch if the C2C_ prefix was missing in the current set value.

Toffer90:
> I've got no idea where pSource data comes from, so could be from exe. Maybe...

The pSource comes from the ini file for sure, but it is only read once by the exe right about when the main menu opens after game launch.

MattOttawa:
Everyone hates ini file settings resetting including me right? Might be nice to use python to read a duplicate ini file that doesn't reset, and use your code to keep resetting the civ4 ini to what's in the duplicate.

Toffer90:
> Everyone hates ini file settings resetting including me right? Might be nice to use python to read a duplicate ini file that doesn't reset, and use your code to keep resetting the civ4 ini to what's in the duplicate.

Huh... It is only one line in the ini file we want to change, and only if it is set wrong, it makes little sense to have a duplicated ini file to change this one value...

If we rename all eras to have a "C2C_" prefix we wouldn't even have to change the value more than once during the first launch of C2C after installing C2C, after that there would be non need to change it again unless someone played a different mod or vanilla before starting C2C again.
Meaning that the last selected era would be rememered like all other custom game option and we would fix it so that prehistoric is the default without having to name it ERA_ANCIENT.

billw2012:
I think the suggestion is to kill two birds with one stone: fixing this specific issue, AND fixing the issue with ini files resetting values occasionally.

Toffer90:
> I think the suggestion is to kill two birds with one stone: fixing this specific issue, AND fixing the issue with ini files resetting values occasionally.

I wasn't aware about the existence of that second issue, has never happened to me.
How would we be able to fix it if it exist though, wouldn't we risk introducing the issue we are trying to fix by adding code that resets the ini file in case it has been mysteriously reset to factory default by the exe... 
The ini file contain a lot of player personalization and would naturally differ from player to player in most ini values.

billw2012:
Yeah not sure, probably such a thing shouldn't be done as part of this fix, as getting it correct might be rather more significant work. There is some tool that copies and restores the civ ini already I think to solve the problem (can't remember the name, just remember it was mentioned when this came up last).

raxo2222:
I just have my civ INI file zipped as backup.

Also I have disabled intro movies in ini, have window screen, and disabled annoying useless popups, so I know well when ini resets for no reason :p

PR #345, opened 2020-01-07
Removed Netbeans.

MattOttawa:
No offence to Ank, hope he shows up again someday.


Toffer90:
I see little reason to delete files that are part of a WIP this soon. What if Ank doesn't keep any of it locally on his machine and come back and see he has to redo stuff. Just saying we should let it be for more than some months before cleaning it out. It doesn't take up resources or impact the game to have it lying there.

billw2012:
> see he has to redo stuff. 

Well source control means you cant get it back easily, nothing is lost. But yeah I can't see any reason to remove these unless they no longer work? We should remove stuff that doesn't work and won't be maintained.

Toffer90:
Yeah, nothing is lost but it takes some energy to dig it up again if it was removed long ago and you don't remember when you added it and don't know when it was removed.

I'm mostly trying to stop a precedence from forming, one where everything that isn't finished yet gets removed a month or two after it was initially added. Some slow modders might get frustrated and disappear if that becomes the norm.

billw2012:
> I'm mostly trying to stop a precedence from forming, one where everything that isn't finished yet gets removed a month or two after it was initially added. Some slow modders might get frustrated and disappear if that becomes the norm.

Well hopefully from now on stuff that isn't finished will just languish in peoples branches and forks so that shouldn't be a problem.

MattOttawa:
Netbeans is a replacement for vs right?

Alberts2:  There's no debugging support for msvc in NetBeans that's a huge disadvantage. 

ofc I can close pull request and forget about it. That wouldn't be a problem.

MattOttawa:
I will open this again at a later date. No rush.

PR #352, opened 2020-01-09
Code Cleanup.

MattOttawa:


Sources/CvPlayer.cpp
```diff
@@ -16687,13 +16687,12 @@ void CvPlayer::setEndTurn(bool bNewValue)
 /*                                                                                              */
 /* Advanced Automations                                                                         */
 /************************************************************************************************/
-			int iLoop;
-			for(CvSelectionGroup* pLoopSelectionGroup = firstSelectionGroup(&iLoop); pLoopSelectionGroup; pLoopSelectionGroup = nextSelectionGroup(&iLoop))
+			foreach_ (CvSelectionGroup* group, groups())
```
billw2012: This is fine how it is, note for future reference you can do this though:
```cpp
foreach_(CvSelectionGroup* group, groups() | filtered(CvSelectionGroup::fn::getAutomateType() == AUTOMATE_SHADOW))
{
	group->setForceUpdate(true);
	group->AI_update();
}
```
(if you implement the fn in the same manner as is done for the CvUnit ones)

Sources/CvPlayer.cpp
```diff
@@ -19146,12 +19100,9 @@ void CvPlayer::updateGroupCycle(CvUnit* pUnit, bool bFarMove)
...
 						MB_OK);
 	}
-	pUnitNode = pPlot->headUnitNode();
 
-	while (pUnitNode != NULL)
+	foreach_ (CvUnit* pLoopUnit, pUnit->plot()->units())
 	{
-		pLoopUnit = ::getUnit(pUnitNode->m_data);
-		pUnitNode = pPlot->nextUnitNode(pUnitNode);
 		//AlbertS2 added this to fix a CTD that happened while loading a saved game
```
billw2012: We should remove things like this, code shouldn't be changed to fix broken saved games like this. It should be constrained to either the loading code or the modifier recalc.

Sources/CvUnit.cpp
```diff
@@ -30860,51 +30807,40 @@ bool CvUnit::canAirBomb1(const CvPlot* pPlot) const
...
+		return false;
 	}
 
-	pCity = pTargetPlot->getPlotCity();
-
+	const CvCity* pCity = pTargetPlot->getPlotCity();
 	if (pCity != NULL)
 	{
-		if (!(pCity->isBombardable(this)))
+		if (!pCity->isBombardable(this))
```
billw2012: This statement can also be moved into the outer if: the right hand side of an && statement is only executed if the left side was true, so you can use it to check for null before using a value in the same statement.

MattOttawa: if (pCity != NULL)
	{
		if (!pCity->isBombardable(this))
		{
			return false;
		}
	}
	else
	{

That won't mess with the else?

billw2012: Yeah you are right I didn't read it properly!

Sources/CvPlayer.cpp
MattOttawa: will do.

Issue #344, opened 2020-01-07
Refinement of Multiple Production feature


raxo2222:
Now you can build multiple wonders per turn as well switch smoothly between food requiring units and buildings.
Code which was commented out also prevented exploits.

https://forums.civfanatics.com/threads/some-more-questions-about-the-mod.653537/page-2#post-15635722

billw2012:
I don't understand your description? "Now you can" sounds like this was fixed? Do you mean "You should be able to"?

billw2012:
Oh okay I didn't notice you did a change. Not sure what this bug is for? I guess you are going to mark it fixed?

raxo2222:
This issue is for exploits being allowed now - you couldn't build multiple wonders in turns or alternate between settlers/workers and buildings.

This will stay open until exploits are fixed while allowing to build multiple wonders per turn, I linked thread here.

billw2012:
Oh okay, I suggest you put in some reference to the code exactly then so we remember when we come back to it. You can link to lines in the files, or just put class + function name.

PR #356, opened 2020-01-09
AI multibuild

billw2012:
AI will build multiple buildings at a time, reducing recalculation requirements.


raxo2222:
I think exploits, that are now opened up should be fixed:
https://github.com/caveman2cosmos/Caveman2Cosmos/issues/344

billw2012:
I don't really think those exploits are a big deal personally, I'm not going to fix them myself any time soon, and this wouldn't be the right place to do it anyway, its a separate issue from multi-build.

PR #358, opened 2020-01-10
Removed MOD_GWARM and _BUFFY #defines.

MattOttawa:
no intended changes to game.


Sources/CvGameInterface.cpp
```diff
@@ -660,14 +660,12 @@ void CvGame::updateColoredPlots()
...
 							{
-// BUFFY - Don't Recommend Plots in Fog of War - start
-#ifdef _BUFFY
-								if (!pLoopPlot->isVisible(pHeadSelectedUnit->getTeam(), false))
-								{
-									continue;
-								}
-#endif
-// BUFFY - Don't Recommend Plots in Fog of War - end
+								// Don't Recommend Plots in Fog of War - start
```
billw2012: Is this the correct fix here? I don't know the context but is the _BUFFY change here an improvement, and if so why not keep it enabled?

Sources/CvUnit.cpp
```diff
@@ -5235,15 +5235,12 @@ bool CvUnit::isActionRecommended(int iAction)
...
-// BUFFY - Don't Recommend Actions in Fog of War - start
-#ifdef _BUFFY
-	// from HOF Mod - Denniz
-	if (!pPlot->isVisible(GC.getGameINLINE().getActiveTeam(), false))
-	{
-		return false;
-	}
-#endif
-// BUFFY - Don't Recommend Actions in Fog of War - end
+	// Don't Recommend Actions in Fog of War - start
```
billw2012: Same question, is it better with this code or not?

MattOttawa: No idea. It's probably never been on. It's up to TB and the others if it's better on or not.

Thunderbrd:
I really have no idea what the buffy mod is all about...

MattOttawa:
Hows a unit supposed to end up in fog of war? Teleport I guess, but then why wouldn't you want that uit selected? You'll never find it. It was probly a bad idea that went nowhere. Maybe.

Thunderbrd:
I thought it might have more to do with suggesting a move into the FOW for scouting and such.

Toffer90:
> I thought it might have more to do with suggesting a move into the FOW for scouting and such.

It's about recommending plots to settle on for the settler unit. It should definitly not reccomend plots that are unexplored (opaque black canvas), but it makes sense to allow it to check fog of war plots.

Thunderbrd:
Might as well remove it then I suppose.

billw2012:
Yeah in that case remove it fully, rather than leave it commented out I think.

PR #359, opened 2020-01-10
Removed 2 unused mothods.

MattOttawa:
Civ buildings and units stuff.
Not used in python.


Toffer90:
Looks good, those two methods would never really be used due to the develpment direction we are taking for civilization unique buildings and units (removal of unitclass and buildingclass plan).

PR #360, opened 2020-01-10
code cleanup.

MattOttawa:


Sources/CvPlot.cpp
```diff
@@ -970,81 +936,62 @@ void CvPlot::doImprovement()
...
 					while (pUnitNode != NULL)
 					{
 						oldUnits.insertAtEnd(pUnitNode->m_data);
 						pUnitNode = nextUnitNode(pUnitNode);
 					}
 
+					bool bDefenderFound = false;
 					pUnitNode = oldUnits.head();
 
 					while (pUnitNode != NULL)
```
billw2012: Possible impl:
```cpp
const bool bDefenderFound = algo::any_of(
    units(), 
    CvUnit::fn::getFortifyTurns() > 0 && CvUnit::fn::getTeam() == getTeam() && CvUnit::fn::canDefend()
);
```

Sources/CvPlot.cpp
```diff
@@ -970,81 +936,62 @@ void CvPlot::doImprovement()
...
+			if ((upgradeInfo.getPrereqTech() == NO_TECH || team.isHasTech((TechTypes)upgradeInfo.getPrereqTech()))
+				&& upgradeInfo.getHighestCost() < GET_PLAYER(getOwner()).getEffectiveGold())
 			{
-				// Super Forts begin *upgrade* - added if-else statement
-				if(GC.getImprovementInfo(getImprovementType()).isUpgradeRequiresFortify())
+				// Super Forts *upgrade*
+				if (GC.getImprovementInfo(eImprovement).isUpgradeRequiresFortify())
 				{
-					bool bDefenderFound = false;
 					CLinkList<IDInfo> oldUnits;
```
billw2012: I can't see the point of this `oldUnits` copy at all can you?

Sources/CvPlot.cpp
```diff
@@ -970,81 +936,62 @@ void CvPlot::doImprovement()
...
-					
 
 					while (pUnitNode != NULL)
 					{
 						oldUnits.insertAtEnd(pUnitNode->m_data);
 						pUnitNode = nextUnitNode(pUnitNode);
 					}
 
+					bool bDefenderFound = false;
 					pUnitNode = oldUnits.head();
```
billw2012: This is the same as `headUnitNode()` isn't it?! Why is it using the `oldUnits.head()` but then using the normal `nextUnitNode` function to get the remaining ones? 
WTF was the author thinking ?!

Sources/CvPlot.cpp
```diff
@@ -16105,21 +15805,14 @@ bool CvPlot::isInUnitZoneOfControl(PlayerTypes ePlayer) const
 {
 	for (int iI = 0; iI < NUM_DIRECTION_TYPES; ++iI)
```
billw2012: FYI There is an range for this, which automatically skips invalid plots: CvPlot::adjacent_range
`foreach_(CvPlot* adjacentPlot, adjacent())`

Sources/CvPlot.cpp
```diff
@@ -16105,21 +15805,14 @@ bool CvPlot::isInUnitZoneOfControl(PlayerTypes ePlayer) const
...
+		const CvPlot* pAdjacentPlot = plotDirection(getX(), getY(), (DirectionTypes)iI);
 		if (pAdjacentPlot != NULL)
 		{
-			CLLNode<IDInfo>* pUnitNode;
-			CvUnit* pLoopUnit;
-
-			pUnitNode = pAdjacentPlot->headUnitNode();
-
-			while (pUnitNode != NULL)
+			foreach_(CvUnit* unit, pAdjacentPlot->units())
```
billw2012: This is better written as something like:
```cpp
if(algo::any_of(pAdjacentPlot->units(), CvUnit::fn::isZoneOfControl()))
{
    return ...
}
```

Sources/CvPlot.cpp
```diff
@@ -16131,17 +15824,14 @@ bool CvPlot::isBorder(bool bIgnoreWater) const
 {
 	for (int iI = 0; iI < NUM_DIRECTION_TYPES; ++iI)
```
billw2012: This can be something like:
```cpp
if(algo::any_of(
    adjacent(), 
    (CvPlot::fn::isWater() && !bIgnoreWater && !isWater()) 
    ||
    (CvPlot::fn::getOwner() != getOwner())
)) 
{
    return true;
}
```

Sources/CvPlot.cpp
```diff
@@ -16456,50 +16145,40 @@ void CvPlot::setImprovementUpgrade()
 		{
 			if ((ImprovementTypes)GC.getBuildInfo((BuildTypes)iI).getImprovement() == eUpgrade)
 			{
-				BuildTypes eBuild = ((BuildTypes)iI);
```
billw2012: You shouldn't remove this, it's better than doing `(BuildTypes)iI` every time. It should be moved up a few lines, made cost, used in all the places.

MattOttawa:
I will join discord once I'm done switching loops and moving variables.

Sources/CvPlot.cpp
```diff
@@ -580,37 +580,13 @@ void CvPlot::updateGraphicEra()
...
+	foreach_(CvUnit* unit, units())
 	{
-		pLoopUnit = ::getUnit(pUnitNode->m_data);
-		pUnitNode = oldUnits.next(pUnitNode);
-
-		if (pLoopUnit != NULL)
-		{
-			pLoopUnit->kill(false, NO_PLAYER, true);
-		}
+		FAssert(unit != NULL);
```
billw2012: Not required, it definitely shouldn't be null ever.

Sources/CvPlot.cpp
```diff
@@ -1636,19 +1557,11 @@ bool CvPlot::isPlotGroupConnectedBonus(PlayerTypes ePlayer, BonusTypes eBonus) c
 
 bool CvPlot::isAdjacentPlotGroupConnectedBonus(PlayerTypes ePlayer, BonusTypes eBonus) const
 {
-	CvPlot* pAdjacentPlot;
-	int iI;
-
-	for (iI = 0; iI < NUM_DIRECTION_TYPES; ++iI)
+	foreach_(CvPlot* adjacentPlot, adjacent())
```
billw2012: This is an `algo::any_of(adjacent(), CvPlot::fn::isPlotGroupConnectedBonus(ePlayer, eBonus))`.

Sources/CvPlot.cpp
```diff
@@ -1759,30 +1649,20 @@ bool CvPlot::shareAdjacentArea(const CvPlot* pPlot) const
...
-				if (pPlot->isAdjacentToArea(iCurrArea))
-				{
-					return true;
-				}
-
-				iLastArea = iCurrArea;
+				return true;
 			}
+
+			int iLastArea = iCurrArea;
```
billw2012: Woah!! That is entirely broken!!!

Sources/CvPlot.cpp
```diff
@@ -1792,44 +1672,22 @@ bool CvPlot::shareAdjacentArea(const CvPlot* pPlot) const
...
-		CvPlot* pAdjacentPlot = plotDirection(getX_INLINE(), getY_INLINE(), ((DirectionTypes)iI));
-
-		if (pAdjacentPlot != NULL && !pAdjacentPlot->isWater())
-		{
-			return true;
-		}
-	}
-
-	return false;
+	return !algo::any_of(adjacent(), CvPlot::fn::isWater());
```
billw2012: There is `algo::none_of` fyi.

Sources/CvPlot.cpp
```diff
@@ -2636,58 +2446,51 @@ void CvPlot::updateSight(bool bIncrement, bool bUpdatePlotGroups)
...
-
-	// Unit
-	while (pUnitNode != NULL)
+	foreach_(CvUnit* unit, units())
 	{
-		pLoopUnit = ::getUnit(pUnitNode->m_data);
-		pUnitNode = nextUnitNode(pUnitNode);
-		
-		if (pLoopUnit != NULL)
+		if (unit != NULL)
```
billw2012: Not needed.

Sources/CvPlot.cpp
```diff
@@ -4106,22 +3904,13 @@ int CvPlot::AI_sumStrength(PlayerTypes eOwner, PlayerTypes eAttackingPlayer, Dom
...
-	CvUnit* pLoopUnit;
-
-	pUnitNode = headUnitNode();
-
-	while (pUnitNode != NULL)
+	foreach_(CvUnit* unit, units())
 	{
-		pLoopUnit = ::getUnit(pUnitNode->m_data);
-		pUnitNode = nextUnitNode(pUnitNode);
+		FAssert(unit != NULL);
```
billw2012: Not needed!

Sources/CvPlot.cpp
```diff
@@ -15803,39 +15373,26 @@ CvUnit* CvPlot::getWorstDefender(PlayerTypes eOwner, PlayerTypes eAttackingPlaye
 
 bool CvPlot::isInUnitZoneOfControl(PlayerTypes ePlayer) const
 {
-	for (int iI = 0; iI < NUM_DIRECTION_TYPES; ++iI)
+	foreach_(CvPlot* adjacentPlot, adjacent())
 	{
-		const CvPlot* pAdjacentPlot = plotDirection(getX(), getY(), (DirectionTypes)iI);
-		if (pAdjacentPlot != NULL)
+		foreach_(CvUnit* unit, adjacentPlot->units() | filtered(CvUnit::fn::isZoneOfControl()))
```
billw2012: This is technically correct but doesn't make it clear what the intent is.
Better would be:
```cpp
bst::optional<CvUnit*> zocUnit = algo::find_if(adjacentPlot->units(), CvUnit::fn::isZoneOfControl());
if(zocUnit)
{
	return GET_TEAM(zocUnit->getTeam()).isAtWar(GET_PLAYER(ePlayer).getTeam());
}
```
However I think the logic is wrong anyway? Surely it should look if *any* zoc units are at war with the player, not just the first one it finds?
Like:
```cpp
if(algo::any_of(adjacentPlot->units(), CvUnit::fn::isZoneOfControl() && CvUnit::fn::isAtWar(ePlayer))
{
	return true;
}
```
(need to add the isAtWar function for units for that one).

Sources/CvPlot.cpp
```diff
@@ -16376,28 +15920,25 @@ int CvPlot::countSeeInvisibleActive(PlayerTypes ePlayer, InvisibleTypes eVisible
 		CvUnit* pLoopUnit = ::getUnit(pUnitNode->m_data);
```
billw2012: Just a loop over units()!

Sources/CvPlot.cpp
raxo2222: What it was doing?

MattOttawa: Meaning I broke it or it was broken before?

billw2012: Its very clear you broke it, just read the code!

Sources/CvPlot.cpp
MattOttawa: Odd debugging code there. 
You see the else? Odd I know.

	foreach_(CvUnit* unit, units())
	{
		if (unit != NULL)
		{
			if (bIncrement)
			{
				unit->changeDebugCount(1);
			}
			else
			{
				unit->changeDebugCount(-1);
			}
			
			changeAdjacentSight(unit->getTeam(), unit->visibilityRange(), bIncrement, unit, bUpdatePlotGroups);
		}
		else
		{
			verifyUnitValidPlot();
			break;
		}
	}

billw2012: Yeah no idea about this code, but unit should never be null in any context!

Sources/CvPlot.cpp
MattOttawa: cause I added int again to iLastArea?

billw2012: Yeah you declared a new variable `iLastArea` and set it to `iCurrArea`.

Sources/CvPlot.cpp
MattOttawa: I added similar FAsserts to CvSelectionGroup. Can I remove them?

Sources/CvPlot.cpp
```diff
@@ -580,37 +580,12 @@ void CvPlot::updateGraphicEra()
...
-	while (pUnitNode != NULL)
-	{
-		oldUnits.insertAtEnd(pUnitNode->m_data);
-		pUnitNode = nextUnitNode(pUnitNode);
-	}
-
-	pUnitNode = oldUnits.head();
-
-	while (pUnitNode != NULL)
+	foreach_(CvUnit* unit, units())
```
billw2012: Should use `units_safe()` here if the units are being deleted in the loop itself, I think they are.

PR #371, opened 2020-01-11
Removing INLINE 2

MattOttawa:
Also removed some tags


billw2012:
Why are you spreading this one change over multiple PRs? Just search replace all the functions for all the code in one go?

PR #374, opened 2020-01-12
Error message when creating user settings folder fails

MattOttawa:


billw2012:
@Toffer90  its okay?

Toffer90:
instead of using CyPythonMgr().errorMsg( I would use a simple "raise" statement, let python handle it with its native tools, the error would pop up as an in-game error message window even, I believe, with default ini settings.
Other than that its ok assuming mat has tested that the message shows up in a good fashion with default ini settings in game when the "try" fails.

MattOttawa:
I tested by adding the message to init and that's all I will do. I've never had had that problem and will not attempt to recreate it.

Toffer90:
> I tested by adding the message to init and that's all I will do. I've never had had that problem and will not attempt to recreate it.

Just add somehing that will intentionally cause an error, like a "abc[4]" line inside the try clause, that will let you check the error message in game, remember to change your ini so that hide python exceptions is set to true first.

Toffer90:
Just tested "raise" and it doesn't show when "hidepythonexceptions = 1"
So forget that suggestion. I'll test with the method you used too mostly just to see how it looks like.

Toffer90:
CyPythonMgr().errorMsg(... soesn't show either with "hidepythonexceptions = 1".
So I'm thinking the message should have to be programmed not as an error but as a regular pop up window akin to any other pop ups in the game.

Toffer90:
Wait, No, this won't work, need to program the message not as an error, but as a pop up window akin to any other in-game UI pop ups.

MattOttawa:
CyPythonMgr is civ4 ingame popup msg.

Toffer90:
Yeah, so is the "raise" statement, though the default ini config does hide the message because is treated as a python error.
I tested you code in game, and I got the error message when I turned off hide python exceptions, but got no message when I turned on hide python exceptions (which is the default state of the ini value)

Toffer90:
With this code 
```py
from CvPythonExtensions import CyPythonMgr
def initUserSettingsDir():
	CyPythonMgr().errorMsg("Error while attempting to create a new user settings folder. This problem can usually be solved by running the mod in administrator mode. Path:/s" + userSettingsDir)
	if not path.isdir(userSettingsDir):
		mkdir(userSettingsDir)
	if not path.isdir(userSettingsDir + "\\DomesticAdv"):
		mkdir(userSettingsDir + "\\DomesticAdv")
```
I got no error message when hidepythonexceptions = 1
I got attached error during game launch when hidepythonexceptions = 0
Untitled-1

PR #375, opened 2020-01-12
Removed 3 unused xml global defines

MattOttawa:


billw2012:
SM is still in development, @Thunderbrd might want those.

Toffer90:
I think TB mentioned that those two SM defines would no longer be needed for his project on the other issue where mat listed a bunch of global defines with no apparent use.

Thunderbrd:
Ah.. yes, those are not needed I don't think.  If they aren't applied now they shouldn't be necessary.

PR #380, opened 2020-01-14
Strength in numbers 1

MattOttawa:


billw2012:
Thanks awesome job, I was about to do this. However it does indicate to me again that we need to be communicating better as we would have ended up doing conflicting changes but by chance. Please join discord already!

Thunderbrd:
Can you safely do that with the read/write wrapper?  Will that not setup all games to become corrupt as soon as strength in numbers is activated by introducing them into the middle of the stream suddenly like that?  Will it not also screw up games NOW by removing them?

MattOttawa:
Removed the write. Turned the read into a wrapper made for this. Dunno how that wrapper works.

MattOttawa:
I removed the write and switched the read to a wrapper that was made by a smarter modder then me that's designed to avoid the problem you mention.
I don't think adding stuff to saves is ever a problem right?
Could make a note in the file that all those SKIP_ELEMENT wrappers can be removed if/when saves are broken.

Thunderbrd:
Adding stuff to saves is a problem if you plan to optionally remove them later or include them later.  Once part of the stream they must remain part of the stream so this is not a toggleable kind of thing.  What you can do is write a proxy variable that picks up the defined value based on whether the tag exists or not and defaults otherwise, then read into a proxy variable and deliver to the actual variable if that variable exists.  This is a super dangerous kind of place to be adding/subtracting variables from the stream so if you can just store a default if the tags aren't initialized and only store what's read if they are, then you should be fine.  Make any sense?

billw2012:
I'm working on changes that make most of this change obsolete (as I discussed with TB already): moving stuff like this in to separate components from the unit itself. So in the unit it would be `UnitComponentStrengthInNumbers* m_sin;`, and we would just not initialize it until the component is complete.

MattOttawa:
Hardy tested.

MattOttawa:
UnitComponentStrengthInNumbers is long.

PR #385, opened 2020-01-15
[WIP] Strength in numbers 2

MattOttawa:
Branched from previous.
Work in progress, wont compile.
Added enums and vectors.


billw2012:
Yes very nice!

MattOttawa:
here's version 2. Next step is to put all the code in the same file the same way bill is doing his unit components.

Thunderbrd:
Improved and I see some of my concerns mentioned elsewhere are off.  However, how are you going to ensure that the write stream writes the same number of entries as the read stream will expect?  I have to admit I'm not sure if that will be correct or not as you have it.  I'm asking because I don't know clearly on that.

PR #389, opened 2020-01-17
Removed Fresh water bad health

MattOttawa:
Not used in python.
This is used when the xml global define for fresh water health is set under 0.


Toffer90:
Yeah, I didn't really commit anything from this PR, though the PR I did commit, removed granular control, had this one in history though all changes from here had been reverted there.

open #389

PR #392, opened 2020-01-18
Rationalized bonus disc/depletion code for improvements

Toffer90:
This should wait till after v40.1 is released, but I hope others will help me by reviewing the code or by testing it in game by turning on the resource depletion BUG option. 
Things to look for when testing which are not related to bonus depletion are  bonus discovery and improvement upgrading as that too was tweaked.

Bonuses can be discovered by culturally owned improvements even if not worked by the citizens of a city.

Mines will no longer perma-deplete, only bonuses may deplete with the resource depletion BUG-option.

Some improvement tags can be removed, but I didn't do this fearing it might break saves.

### DIFF
--- a/Assets/XML/Art/CIV4ArtDefines_Improvement.xml
+++ b/Assets/XML/Art/CIV4ArtDefines_Improvement.xml
@@ -467,7 +467,7 @@
 			<KFM>Art/Structures/Improvements/OilWell/OilWell.kfm</KFM>
 			<Button>Art/Interface/Buttons/Builds/Groundwater_well.dds</Button>
 		</ImprovementArtInfo>
-		<ImprovementArtInfo>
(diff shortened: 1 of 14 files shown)


raxo2222:
Why if (!isDisorder()) was deleted in cvcity.cpp?

MattOttawa:
Part of:
Bonuses can be discovered by culturally owned improvements even if not worked by the citizens of a city.

Toffer90:
> Why if (!isDisorder()) was deleted in cvcity.cpp?

Because doImprovement() which was the only thing inside that clause was moved to another place.
It was moved from CvCity::doTurn() to CvPlot::doTurn(). 
Additionally, I saw no reason why city disorder should affect the improvements of the countryside.

PR #401, opened 2020-01-21
Removed all usage of MAX_CIV_X from python and dll; and improved some bad rev code.

Toffer90:
Should wait till after v40.1
The changes are probably sound.
I did 200 turns AI autoplay on blitz gamespeed without issues.
The python logs said revolution code had performed a couple revolts and revolutions during those 200 turns.


raxo2222:
There is MAX_TEAMS and MAX_PLAYERS too.

Toffer90:
> There is MAX_TEAMS and MAX_PLAYERS in python too.

What's your point?

raxo2222:
> 
> 
> > There is MAX_TEAMS and MAX_PLAYERS in python too.
> 
> What's your point?

I'm not sure if they are used correctly in all places in python

MattOttawa:
I bet this is a step forward either way.

Toffer90:
> I'm not sure if they are used correctly in all places in python

Found two places it seemed wrong to use it, though my focus with this PR was mostly to isolate MAX_CIV_X so that they are only used by the exe. Enabling us to experiment a bit with them to see how the exe reacts without having to worry about messing up python and dll at the same time.

PR #406, opened 2020-01-29
const

MattOttawa:


Sources/CvBuildingInfo.cpp
```diff
@@ -2231,12 +2231,12 @@ bool CvBuildingInfo::isPrereqAndCivics(int i) const
 }
 
 //This is for the readpass3
-int CvBuildingInfo::isPrereqOrCivicsVectorSize()					{return m_aszPrereqOrCivicsforPass3.size();}
-CvString CvBuildingInfo::isPrereqOrCivicsNamesVectorElement(int i)	{return m_aszPrereqOrCivicsforPass3[i];}
-int CvBuildingInfo::isPrereqOrCivicsValuesVectorElement(int i)		{return m_abPrereqOrCivicsforPass3[i];}
-int CvBuildingInfo::isPrereqAndCivicsVectorSize()					{return m_aszPrereqAndCivicsforPass3.size();}
-CvString CvBuildingInfo::isPrereqAndCivicsNamesVectorElement(int i)	{return m_aszPrereqAndCivicsforPass3[i];}
-int CvBuildingInfo::isPrereqAndCivicsValuesVectorElement(int i)		{return m_abPrereqAndCivicsforPass3[i];}
+int CvBuildingInfo::isPrereqOrCivicsVectorSize() const					{return m_aszPrereqOrCivicsforPass3.size();}
```
billw2012: Looks like alignment messed up on a couple of these lines (remember github tab size is not 4 by default, check it in a proper editor).

Sources/CvGlobals.h
```diff
@@ -682,61 +682,61 @@ class cvInternalGlobals
 /* Afforess                                END                                                  */
 /************************************************************************************************/
 
-	int getNumCityTabInfos();
+	int getNumCityTabInfos() const;
 	std::vector<CvInfoBase*>& getCityTabInfos();
-	CvInfoBase& getCityTabInfo(CityTabTypes e);
+	CvInfoBase& getCityTabInfo(CityTabTypes e) const;
```
billw2012: I think all these functions should return const references, doubly so if the object itself is const. We shouldn't ever be modifying the CvInfo classes by definition right? We do it during loading I think, so we need some way to do it though...

Sources/CvInfos.h
```diff
@@ -232,22 +232,22 @@ class CvDiplomacyResponse
 	CvDiplomacyResponse();
 	virtual ~CvDiplomacyResponse();
 
-	int getNumDiplomacyText();
+	int getNumDiplomacyText() const;
 	void setNumDiplomacyText(int i);
 
-	bool getCivilizationTypes(int i);
+	bool getCivilizationTypes(const int i) const;
```
billw2012: Its not really idiomatic to make integral type parameters const (int, float, char), only references and pointers. It isn't wrong per se, but you will have an uphill battle against the rest of the C++ community to get them to do it 🙂

Sources/CvPlayerAI.cpp
```diff
@@ -4357,12 +4357,12 @@ CvCity* CvPlayerAI::AI_findTargetCity(CvArea* pArea) const
...
 
 
-bool CvPlayerAI::AI_isCommercePlot(CvPlot* pPlot) const
+bool CvPlayerAI::AI_isCommercePlot(const CvPlot* pPlot) const
 {
 	return (pPlot->getYield(YIELD_FOOD) >= GC.getFOOD_CONSUMPTION_PER_POPULATION());
 }
 
-bool CvPlayerAI::AI_getVisiblePlotDanger(CvPlot* pPlot, int iRange, bool bAnimalOnly, CvSelectionGroup* group, int acceptableOdds) const
+bool CvPlayerAI::AI_getVisiblePlotDanger(const CvPlot* pPlot, int iRange, bool bAnimalOnly, CvSelectionGroup* group, int acceptableOdds) const
```
billw2012: `group` should be const as well surely?

Sources/CvPlayerAI.cpp
```diff
@@ -12744,7 +12744,7 @@ int CvPlayerAI::AI_unitPropertyValue(UnitTypes eUnit, PropertyTypes eProperty) c
 /* 	City Defenders						24.07.2010				Fuyu			*/
 /********************************************************************************/
 //Fuyu bIgnoreNotUnitAIs
-int CvPlayerAI::AI_unitValue(UnitTypes eUnit, UnitAITypes eUnitAI, CvArea* pArea, CvUnitSelectionCriteria* criteria) const
+int CvPlayerAI::AI_unitValue(UnitTypes eUnit, UnitAITypes eUnitAI, const CvArea* pArea, CvUnitSelectionCriteria* criteria) const
 /********************************************************************************/
 /* 	City Defenders												END 			*/
 /********************************************************************************/
```
billw2012: `criteria` should be const as well?

Sources/CvBuildingInfo.cpp
MattOttawa: Those will be moved to the header files according to your plan. I'll make sure it's all good if I do it.

Sources/CvGlobals.h
MattOttawa: Step one is to add as much const to CvInfos as possible. (other pr has a lot)
I'll continue this as soon as I can.

Sources/CvPlayerAI.cpp
MattOttawa: I'll get these 2 in another pr.

PR #407, opened 2020-01-29
2 namespace functions

MattOttawa:


Sources/CvGame.cpp
```diff
@@ -176,11 +235,11 @@ void CvGame::init(HandicapTypes eHandicap)
 
 		for (int i = 0; i < iNumPlayers; i++)
 		{
-			int j = (getSorenRand().get(iNumPlayers - i, NULL) + i);
+			const int j = (getSorenRand().get(iNumPlayers - i, NULL) + i);
 
 			if (i != j)
 			{
-				int iTemp = aiTeams[i];
+				const int iTemp = aiTeams[i];
```
billw2012: FYI There is a function `std::swap<>()` for this for future reference.

Sources/CvGame.cpp
```diff
@@ -118,6 +118,65 @@ CvGame::~CvGame()
...
+	{
+		for (int iI = 0; iI < NUM_GAMEOPTION_TYPES; iI++)
+		{
+			const GameOptionTypes eGameOption = static_cast<GameOptionTypes>(iI);
+			if (game->isOption(eGameOption))
+			{
+				CvGameOptionInfo& optionInfo = GC.getGameOptionInfo(eGameOption);
+
+				const int numOptionsForcedOn = optionInfo.getNumEnforcesGameOptionOnTypes();
+				if (numOptionsForcedOn > 0)
```
billw2012: This `if` is pointless (and the one below it), if the value is 0 the loop won't do anything anyway.

billw2012:
So I said "its very nice" and then you closed it with out comment?!

PR #408, opened 2020-01-31
Current era cache

MattOttawa:
Needs a tweak to make it saved game compatible.


MattOttawa:
Just gotta test, but I don't like testing much. I'll do it later. Easy to reopen.

PR #409, opened 2020-01-31
Removed granular callback controller

MattOttawa:
Maybe somebody can make a note of when this pr is merged so modders can retrieve it later. (if it's merged)


Toffer90:
This file can be removed as part of this PR related to the granular callback controller.
AnimalPlacing_PythonCallbackDefines.xml
It is not used in any way even without this branch PR, it only exist as an example on how the xml part would look like.

PR #410, opened 2020-01-31
Added REMAPPED_CLASS_TYPE code for yields, commerce and domains

MattOttawa:


MattOttawa:
#99

billw2012:
This just adds the ability for us to do the remapping if we want, but doesn't actually change any existing behaviour correct?

MattOttawa:
correct.

PR #411, opened 2020-01-31
[PM] Finished exposing a few functions to python

MattOttawa:

PR #412, opened 2020-01-31
[PM] updated map generator

MattOttawa:
just trying to spam the github feed. ha. I know people miss me.

If anyone is paying attention, the following prs are part of the same modification, but I will be making separate draft prs because when you put it all together the end result is an infinite loop I think. 
This is the first bit of c++ I ever done, so it took months of messin around to get this far and I'm done with it. This stuffs also been sittin around for a year, so I gotta do somethin with it. Here's someone else's chance. 

Thanks to LyTning94, AIAndy, and Koshling for gettin things started.


MattOttawa:
Instructions for modders:
#414 - The python is needed.
#411 - exposed functions is needed, so I merged it into the python.
#415 - CvMap fix is also needed I think but not merged.
Testing should be possible with those 3.

PR #413, opened 2020-01-31
[PM] updated player members

MattOttawa:
finale copying from LyTning94.


MattOttawa:
This is not saved game compatible the way it is.

PR #416, opened 2020-02-01
FIX(Pedia) - Bonus XML file + typos

SannitaSSJ:
Bonus XML file: re-sorting alphabetically, removing duplicates, fixing/removing wrong lines and tags, adding/fixing Italian translations; other files: fixing wrong names ("Xiongu" --> "Xiongnu"; "Yanomani" --> "Yanomami") and several typos


SannitaSSJ:
Ok, nevermind, I'll re-try with separating the various requests

PR #420, opened 2020-02-03
added foreachPlot() to CvMap

MattOttawa:
skipped some spots cause I don't know how to use bst::bind yet.


MattOttawa:
I'll probly redo this using algo.

PR #421, opened 2020-02-05
manySmallChanges

Toffer90:
I went slow with the changes to not make silly mistakes, I've also tested each change to see that they work as intended. I believe it is safe enough and worth it  to go in before v40.1, but I wouldn't want to upset anyone by oushing it straight to master. Probably best to hear what opinions the rest of the team has about this. I wouldn't mind if it were postponed until after release as there's nothing here that is critical in any way. One city challenge may still have other bugs than the ones I fixed here, I took care of the obvious AI exceptions to the rule only.

Improved new game starting unit code.
e.g. Wrub of the neanderthals now start prehistoric games with neanderthal units instead of sapiens.

One city challenge (OCC) now apply to AI too.

Redefined the xp values set for units in xml from how much other units get from killing me to how much I get for killing other units (one value for defense and another for offense), makes a lot more sense that way, Gives us the ability to define how fast a specific unit improves through defensive and offensive victories, heroes should have low base xp values as they are already strong from the get go, wild animals can now get more exp from winning battles regardless of what they win against. XP amounts are anyway adjusted by the difficulty of battles.

Added Neanderthal Tribal Guardian unit, removed a dead handicap tag, misc clean up.


MattOttawa:
better for conflicts if you merge.

raxo2222:
It fails to compile with final release build

Toffer90:
@raxo2222 Now stop making the branch more extensive than it already is, any commits to branches should be to fix issues in the branch, not change for change's sake.

raxo2222:
> 
> 
> @raxo2222 Now stop making the branch more extensive than it already is, any commits to branches should be to fix issues in the branch, not change for change's sake.
> 

Well there were more tags, that were marked as unused.

> > It fails to compile with final release build
> 
> Final release compiled fine for me.

As I wrote later it compiled after deleting Build folder.

Toffer90:
> As I wrote later it compiled after deleting Build folder.

You also wrote "As if this folder needs to be purged (maybe except deps folder in it) every time you compile DLL."
I only said that I didn't have to delete this folder for it to compile final release.

Toffer90:
@Thunderbrd @billw2012 any opinion regarding this and v40.1 release?

Thunderbrd:
Sounds like it could go in.  Do it now if you're going to so that it gets a little playtesting before 30.1

PR #425, opened 2020-02-07
stuff I missed when removing Python modules info

MattOttawa:


Toffer90:
All good as far as I can deduce, python doesn't break in game in any way from these python specific clean up.

PR #426, opened 2020-02-10
Removed barbarian trait + minor cleanup

raxo2222:
This barbrian trait was unnecessary - it was here when barbarians were weaker and couldn't accumulate culture.
Also it can be turned off like any trait with options.

There were also boosts to culture and food production for NPCs.


Toffer90:
We should keep the trait and remove the native culture building instead.
That is what makes most sense. Less autobuild buildings to process each turn and traits are less memory intensive than buildings are as they have far fewer variables cached.

raxo2222:
I remember that @Thunderbrd wanted to remove some if not most of boosts from this trait.
Maybe is it possible to always have it - just like Ancient Ways buildings that give traits.

Toffer90:
> I remember that @Thunderbrd wanted to remove some if not most of boosts from this trait.
> Maybe is it possible to always have it - just like Ancient Ways buildings that give traits.

Only some of the tags used in the trait have zero impact. Like the number of free unit bonus, barbs don't have economy, nor do they strike.

PR #427, opened 2020-02-10
CvPlayer algo

MattOttawa:


Sources/CvPlayer.cpp
```diff
@@ -6588,43 +6542,15 @@ int CvPlayer::countCityFeatures(FeatureTypes eFeature) const
...
-	{
-		if (pLoopCity->getNumBuilding(eBuilding) > 0)
-		{
-			iCount += pLoopCity->getNumBuilding(eBuilding);
-		}
-	}
-
-	return iCount;
+	return algo::accumulate(cities()
+		| filtered(CvCity::fn::getNumBuilding(eBuilding) > 0)
```
billw2012: The filter is pointless unless `getNumBuildings` can be less than 0. In fact it is worse than pointless as it costs way more to check the value then branch than it does to just add 0 to something! The filter should be removed!

PR #431, opened 2020-02-12
Simplified the NPC setup.

Toffer90:
Mostly code refactoring than anything else at this point.


Thunderbrd:
I would doubt that it would be save compatible if you are adding to the total number of NPCs between one save to the next.  The main reason being that during the read wrapper, you're cycling through a null player that has never been initialized.  Bill might be able to figure out how to manage that.  There's also the exe to worry about in that it might have some hangups about that and I'd be surprised if it didn't.  That's navigable by making sure you have some NPC slots set aside, at least enough that for the foreseeable future, we probably won't need more than that.

Toffer90:
> I would doubt that it would be save compatible if you are adding to the total number of NPCs between one save to the next. The main reason being that during the read wrapper, you're cycling through a null player that has never been initialized.

I think it should be quite possible to, before reading in the NPC data from the sream to check if the stream actually contains another NPC, and if it doesn't initialize the NPC instead of reading in its data from the stream.

Thunderbrd:
That was what I was thinking Bill might be able to figure out how to do... but that's really only possible if the exe will work with that, and there could be thing wrapped up in the way the wrapper stream works that may make that impossible... I don't know.

Sources/CvPlayerAI.cpp
```diff
@@ -24495,20 +24478,14 @@ void CvPlayerAI::read(FDataStreamBase* pStream)
...
-	//		//GET_TEAM(eTeam).init(eTeam);
-	//		//GC.getInitCore().setTeam(ePlayer, eTeam);
-	//	}
-	//}
+	if (getID() == MAX_PC_PLAYERS)
+	{
+		//Read NPC data
+		for (int iI = MAX_PC_PLAYERS+1; iI < MAX_PLAYERS; iI++)
+		{
+			GET_PLAYER((PlayerTypes)iI).read(pStream);
```
MattOttawa: It's just this read that breaks saves right? There's a C2C version that can be used to skip this I think.

MattOttawa: nevermind. Didn't read before posting.

Toffer90: > It's just this read that breaks saves right? There's a C2C version that can be used to skip this I think.

No the read doesn't break saves it is entirely necessary for any NPC units or cities that existed in a save to exist after loading the save.
What makes this branch break old saves is that it changes the enumeration values for all the current NPC teams.

Toffer90:
Ok, so I noticed that bug where the exe doesn't update the graphics of city bars for NPC's above barbarian, so I gotta rethink this one quite a bit.

Toffer90:
If the exe asks for the value of MAX_CIV_PLAYERS more than once, one might get away with changing it when inside a game to MAX_PLAYERS-1 and back to MAX_PC_PL:AYERS when returning to main menu.

Thunderbrd:
> Ok, so I noticed that bug where the exe doesn't update the graphics of city bars for NPC's above barbarian, so I gotta rethink this one quite a bit.

Ah yeah, that was the thing I couldn't get around with the barbs not being last.  Forgot about that one.

Toffer90:
> > Ok, so I noticed that bug where the exe doesn't update the graphics of city bars for NPC's above barbarian, so I gotta rethink this one quite a bit.
> 
> Ah yeah, that was the thing I couldn't get around with the barbs not being last. Forgot about that one.

Yeah, it proved impossible to get around.

Thunderbrd:
> > > Ok, so I noticed that bug where the exe doesn't update the graphics of city bars for NPC's above barbarian, so I gotta rethink this one quite a bit.
> > 
> > 
> > Ah yeah, that was the thing I couldn't get around with the barbs not being last. Forgot about that one.
> 
> Yeah, it proved impossible to get around.

Right, because somehow that city display is handled in a loop inside the exe is what I concluded because I couldn't find anywhere in our assets that indicated anything like it, which was then to say that it was reading directly from one of those established defines, which was ALSO then to say it expects those defines to BE defined at load time, rather than something that can adjust in-game.  Sorry I couldn't remember what it was I found was insurmountable there to let you know before spending your time on all that, but I'm glad you were able to try.  That shows a lot of increasing understanding of the code so that's very cool.  I knew you were going to find something I was forgetting about.  I really tried so many ways to do things here and only gave up when I came to the conclusion I had no better option.  Who knows... maybe Bill might find something is more possible here than I could.

So if I'm not mistaken, this pretty much is the end-all explanation of what forced me back around to the way it's all being done now, the wall I couldn't overcome and why I could also not fix the potential player introduced bug in the staging room.

Toffer90:
>> Yeah, it proved impossible to get around.
>Right, because somehow that city display is handled in a loop inside the exe is what I concluded 
So if I'm not mistaken, this pretty much is the end-all explanation of what forced me back around to the way it's all being done now, the wall I couldn't overcome and why I could also not fix the potential player introduced bug in the staging room.

Yeah the number provided to the exe from "int cvInternalGlobals::getMaxCivPlayers() const" at game launch is the number it use to decide how many players can be selcted in the staging room, but is also the number it use to loop through all the players with when updating city art for their cities.

This branch still simplifies the NPC setup by not having a strange gap between the last real player slot and the first NPC slot. A gap that is messy to deal with when one is looping through all players and 1 NPC and then lopping back to player 0 and up to the one you started with. But that is also pretty much all this branch does now.

PR #437, opened 2020-02-15
Internal tag renames - savebreaking changes.

raxo2222:
Saves will be broken, as they rely on internal tag names.
This will simplify modding, as those tags will match what you see in game.


raxo2222:
So far now all ERA_, CIVICOPTION_, CIVIC_ and RELIGION_ tags match what you see in game.

raxo2222:
Promotions, buildings and units will wait until removal of building/unit classes and after TB finishes unit review.

PR #439, opened 2020-02-15
Fixed crash

MattOttawa:


MattOttawa:
team can fix this when they realize it's broken.

PR #446, opened 2020-02-21
removed building class from the dll

MattOttawa:
Don't remember what I done here. I need the xml done to see if I broke everything.
I do python next.


raxo2222:
Game loaded just fine, but there was compile warning.
Also got python error in pedia when I clicked on tech, same with building and units.
Map generated properly, turns progresses fine.
I could even load that save.

raxo2222:
Game loaded just fine, but there was compile warning.
Also got python error in pedia when I clicked on tech, same with building and units.
Map generated properly, turns progresses fine.
I could even load that save. There is lots of asserts though.

MattOttawa:
I fixed the xml and added readpass2.
Just waitin for stuff to load.

MattOttawa:
i also switched isWorldWonderClass() calls to isWorldWonder() on the dll.

Toffer90:
> WonderClass

Of course, I should have figured that you had renamed those already. ^^
I'll fix that in python right now.

PR #452, opened 2020-02-26
Remove unnecessary IFDEFs

raxo2222:


MattOttawa:
#ifdef means code will be on.
#ifndef means code will be off, so code should be commented out or removed when #ifndef is removed.

Toffer90:
You've reenabled some code that was supposed to be disabled, like this in CvCityAI
```cpp
#ifndef C2C_BUILD
	// if we need to pop borders, then do that immediately if we have drama and can do it
	// cppcheck-suppress knownConditionTrueFalse
	if ((iTargetCulturePerTurn > 0) && (getCultureLevel() <= (CultureLevelTypes) 1))
	{
		if (AI_chooseProcess(COMMERCE_CULTURE))
		{
			return;
		}
	}
#endif
```
"#ifndef X" = "if not def X", so any code inside it must be removed if you remove the define.

Toffer90:
Same thing here`
```cpp
//TB Build Mod (Moves below to lower priority)
#ifndef C2C_BUILD
	if	(!bLandWar && !bAssault && (iTargetCulturePerTurn > getCommerceRate(COMMERCE_CULTURE)))
	{
		if (AI_chooseBuilding(BUILDINGFOCUS_CULTURE, bAggressiveAI ? 10 : 20, 0, bAggressiveAI ? 33 : 50))
		{
			if( gCityLogLevel >= 2 ) logBBAI("      City %S uses minimal culture rate", getName().GetCString());
			return;
		}
	}
#endif
```

Toffer90:
And here
```cpp
//TB Build Mod (Move up in priority)
#ifndef C2C_BUILD
	if (AI_chooseBuilding(BUILDINGFOCUS_PRODUCTION, 30, 10 / iMinValueDivisor))
	{
		return;
	}
#endif
```

Toffer90:
Ok, checked all changes with C2C_BUILD define and you only made that mistake three times in CvCityAI.cpp

PR #460, opened 2020-03-02
Tweaks to bonus filters, Starsign fixes and barbarian nerf

raxo2222:


Toffer90:
Just confirming that it seems you cherrypicked starsigns out of CvEventManager.py correctly.

PR #470, opened 2020-03-07
Switched an if statement to a FAssert

MattOttawa:
I haven't tested yet.


Toffer90:
This change is fine as long as GC.getInitCore().getOption(eIndex); accepts out of bound inputs and return false if it's the case.

Toffer90:
Had a look, and the same assert is inside `CvInitCore::getOption(...)` pretty much.
```cpp
bool CvInitCore::getOption(GameOptionTypes eIndex) const
{
	int numGameOptionTypes = NUM_GAMEOPTION_TYPES;
	FASSERT_BOUNDS(0, numGameOptionTypes, eIndex, "CvInitCore::getOption");
	FAssertMsg(m_abOptions != NULL, "Access to unconstructed game option array");
```

Toffer90:
It's a bit redundant to have basically the same assert two lines in a row.

Toffer90:
You can actually remove the "else" in bool CvInitCore::getOption(GameOptionTypes eIndex) const
Because if the "if" statement is true it won't reach whatever is after the if due to the return within it.

Doesn't matter one way or another, I'm just voicing my subjective opinion on how I would like it.

MattOttawa:
For that kinda stuff I like return cond ? a : b; 
Extra else I like the least.

Toffer90:
> For that kinda stuff I like return cond ? a : b;
> Extra else I like the least.

Yeah, that's also nice; I'm so used to python where that is not possible without creating a helper function that would be slower than the if . . . else . .

PR #493, opened 2020-03-26
Global defines vector

MattOttawa:
Added a vector for int global defines that is accessed using an enum.
Added macros that I copied from AdvCiv by f1rpo.


alberts2:
You merged this before i had the chance to ask what are the goals of this change?

MattOttawa:
I'm making an enum for each global define.
I plan on switching all the current getDefineINT calls to use an enum instead of searching by name.
I also plan on switching the cached versions in CvGlobals.

alberts2:
> I also plan on switching the cached versions in CvGlobals.

Getting the values of those defines will become slower as it is now.

MattOttawa:
eh? Well I thought about making them all that way.... With the macro I can hide all the code but it still needs to be compiled, and I wanna add all the global defines (hundreds).
I can forget about the vector and make a get function for each one.

alberts2:
Getting a value from a vector is always slower than returning a variable. Also keep in mind that global defines can be changed while the game is running.

PR #500, opened 2020-03-27
Switched what I was doing from enums and a vector to separate vars and functions

MattOttawa:


Sources/CvGlobals.cpp
```diff
@@ -467,11 +316,10 @@ cvInternalGlobals::cvInternalGlobals()
 	, m_iStoreExeSettingsCorporationInfo(0)
 	, m_iStoreExeSettingsBonusInfo(0)
 	, m_bSignsCleared(false)
+#define ADD_TO_CONSTRUCTOR(VAR) \
```
MattOttawa: I may need to add a second parameter to the macro so that I can give different starting values. But these values should always get over-ridden... so.... idk.

alberts2:
I try to look at and test these changes next weekend.

PR #505, opened 2020-03-30
removed 3 get functions from CvGlobals.

MattOttawa:
getCivPlayers() shouldn't be used in the dll or python anymore right?
I didn't change the values being sent to the exe. I know toffer just sorted this stuff out.


Toffer90:
Yeah, no need to use those get functions in dll or python, we have the MAX_PC_PLAYERS macro to use instead.

Issue #497, opened 2020-03-27
Remove unused Python exports


ghost:
Currently lots of CvGameCore.dll functions which are exposed to python are not used in any python code. This adds lots of unnecessary source code and increases the CvGameCore compilation time.

Toffer90:
They are easy enough to add back one by one when they are needed, so most unused ones could be removed imo.

MattOttawa:
Some Classes have a Cy counterpart and others are "simple enough to be exposed directly" with just the interface file. Not having the .h and .cpp files would reduce the amount of code without limiting the python. Calls from python don't have the extra step of going to the Cy class first. Exposing new functions is also easier.
It seems to me like the Cy class is needed to send a dll object to python as a parameter for a callback function (any sent to the EntryPoints and the calls to mapscripts . I can't really think of anything else that makes it 100% necessary.
If I'm not wrong, then there are a couple classes that don't need thier Cy class. Players and Teams are never sent to python, only thier enums are.

Toffer90:
Maybe you meant something else, but:
GC.getActivePlayer returns a CyPlayer object....
GC.getPlayer(ePlayer) returns a CyPlayer object....
GC.getTeam(eTeam) returns a CyTeam object....
... to python

MattOttawa:
GC.getPlayer(ePlayer) can return CvPlayer the same way CvInfos and CvProperties are done. I think?

Toffer90:
Maybe, I have no idea really.

MattOttawa:
Actually python doesn't use enums much like the dll. That might be an issue.

MattOttawa:
When you call getFirstUnit() or whatever on CvPlayer it will return CvUnit, so you could change the dll side's name of the function in CyPlayerInterface to pyGetFirstUnit or somethin and copy/paste the code from CyPlayer into CvPlayer. But if that's done for too many things, then it defeates the purpose.

MattOttawa:
There's a lot I need to learn about classes.  Here's idea anyways.
Is it possible to make the cy class a derived class of it's corresponding cv class, then expose functions in the interface file using the cy class name but not have that function in the cy class (virtual or otherwise), so that the function will be handled by the cv base class?
That way adding functions to the cy classes can be skipped if nothing needs to be done
like switching enums to ints or cv classes to cy.

PR #512, opened 2020-03-31
removing unused functions from Cy files

MattOttawa:


MattOttawa:
#497

Toffer90:
You've removed a function called from C2C_Pangaea.py
.def("setPlotTypes", &CyMapGenerator::setPlotTypes, "void (list lPlotTypes) - set plot types to the contents of the given list")

That's the only checked CyMapGenerator, all other removals in CyMapGenerator was ok.

Toffer90:
.def("getNumBonusesOnLand", &CyMap::getNumBonusesOnLand, "int () - total bonuses on land plots")
.def("resetPathDistance", &CyMap::resetPathDistance, "void ()")
.def("calculatePathDistance", &CyMap::calculatePathDistance, "finds the shortest passable path between two CyPlots and returns its length, or returns -1 if no such path exists. Note: the path must be all-land or all-water")

Called from perfectWorld, perfectMongoose and world mapscript.

MattOttawa:
I forgot to search map scripts.

PR #515, opened 2020-04-02
forEachPlot function

MattOttawa:


Toffer90:
Won't this be a reduction in performance without much gain in readability
e.g.
From:
```cpp
for (int iI = 0; iI < numPlots(); iI++) plotByIndex(iI)->erase();
```To:
```cpp
forEachPlot(this, CvPlot::erase()) { for (int iI = 0; iI < map->numPlots(); iI++) map->plotByIndex(iI)->func;;}
```

MattOttawa:
Not exactly. A macro writes stuff at the start of compiling. There will be no function called forEachPlot in the compiled dll.
Still not much of an enhancement in any way.
Here's a comment i removed in this pr:  // XXX generalize these funcs? (macro?)
I made a macro but I havn't generalized nothin.
I'm thinkin I gotta replace the  unique functions in CvMap with forEachPlot wherever they are called and get rid of the unique functions.

Toffer90:
Ah ok, the compiler will compile it pretty much the same either way; it's ok then.
It does make the code a little more compact so it's not entirely for nothing.

alberts2:
Personally i would never use Macros just for eye candy stuff like these loops. Because they have lot's of disadvantages like you can't step into them with the debugger.

Toffer90:
Debugging is important... once again I have objections to this PR.
Bad idea → ok idea → bad idea.

PR #517, opened 2020-04-03
civ mega pack

MattOttawa:


raxo2222:
I think you should start with original one https://forums.civfanatics.com/threads/c2c-civilizations-megapack.538110/

Toffer90:
We shouldn't have this many civs in core, it clutters the civ selection list, increases launch time (especially the leaderhead art); it is just more practical as a modmod.

raxo2222:
He could be just fixing it on github

Toffer90:
> He could be just fixing it on github

Then why make a pull request to master?

MattOttawa:
miccal2000 is sayin xml validator won't work for linux. This will work on linux.

raxo2222:
It could be put to Unloaded Modules too

MattOttawa:
Based on public opinion, I like if someone fixes this, then adding in to unloaded modules makes sense.

Toffer90:
All the art files would still be loaded though. Would be about 1 MB per 10 leaders.
How many civs are in this pack? I'm guessing most of them have two leaders each.

PR #532, opened 2020-04-08
Emerging minor civ golden age

MattOttawa:
Suggested by DH.
I included a global define to set number of turns (currently 1).

This still needs a test.
I'll probly ask players for a save and hope for the best rather then messin with WB.


Toffer90:
Number of turns should scale with gamespeed with 1 as minimum value after scaling.
The global would be number of turns on Normal gamespeed.

alberts2:
Maybe the Golden Age lenght should also be modified by all the other modifiers not just the gamespeed.

PR #535, opened 2020-04-09
Algo in one spot

MattOttawa:
Noticed a spot that can be switched to algo while looking at toffer's stuff.
Not important.
I'm just trying to find as many spots as I can and they pop out at me now.


Toffer90:
This PR is a duplicate, no need for it to exist... PR #534 (https://github.com/caveman2cosmos/Caveman2Cosmos/pull/534) is the original which contains all this PR contains.

PR #537, opened 2020-04-11
City Plot Iterator

alberts2:
Needs to be tested.

At the moment it is only used in `CvCity::calculateFeatureHealthPercentChange`
@MattOttawa 
Maybe you could implement this Iterator at other places as well.


Toffer90:
Good stuff.

PR #539, opened 2020-04-12
Update for an old project

MattOttawa:
Figured out the infinite loop. It happens if code is run before first turn is ended, so here's a little more done.
alt + 1 after first turn to run.


MattOttawa:
I'm going to break this up and put what I can in the mod even though it's incomplete.
It's too big to keep the whole project on a branch and the first part is already in the mod anyways.

PR #541, opened 2020-04-13
removed 4 more functions from CvGlobals.

MattOttawa:
They are each a second variation that isn't needed.


MattOttawa:
hmm... I screwed this up trying to merge master branch. Easier to remake this tiny pr.

PR #547, opened 2020-04-18
Added back the assert for getInfoTypeForString calls that come from python

MattOttawa:
Id like to go a step further and get rid of the ability to hide asserts for getInfoTypeForString altogether.
It doesn't make sense to have references to xml types that don't exist.

### DIFF
--- a/Sources/CyGlobalContext.cpp
+++ b/Sources/CyGlobalContext.cpp
@@ -462,7 +462,7 @@ CvPropertyInfo* CyGlobalContext::getPropertyInfo(int i) const
 
 int CyGlobalContext::getInfoTypeForString(const char* szInfoType) const
 {
-	return GC.getInfoTypeForString(szInfoType, true);
+	return GC.getInfoTypeForString(szInfoType);
 }
 /************************************************************************************************/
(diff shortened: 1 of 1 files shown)


Toffer90:
I agree, it should throw an assert when py asks for the index of something that doesn't exist.

alberts2:
This breaks the modular xml loading.
Because if you turn of a optional module some types are missing that's not a problem because 'getInfoTypeForString' returns '-1' for missing types. The code referencing any types has to be able to handle missing types.

alberts2:
> I agree, it should throw an assert when py asks for the index of something that doesn't exist.

In this case python code should check the return value of `getInfoTypeForString`. If it returns `-1` something is missing and the python code should be programmed to handle this situation.

MattOttawa:
There is a method called getInfoTypeForStringWithHiddenAssert that is in use for some modular civs in python. Modular stuff should use that call instead of removing assert for everything.

alberts2:
`getInfoTypeForStringWithHiddenAssert` is way to long for a function name. Maybe the `getInfoTypeForString` in python could be changed to have a second parameter to disable the assert like it does in c++.

PR #553, opened 2020-04-19
Added bHideAssert parameter to getInfoTypeForString in CyGlobalContext.

MattOttawa:
Removed getInfoTypeForStringWithHiddenAssert.


MattOttawa:
This breaks the python. It seems like the default for the new parameter doesn't work.

alberts2:
I think it's not possible to get use default parameters in python exposed functions.

But having to use that second parameter should not be a problem. The programmers using that function just need to decide if the assert should be hidden or not each time they use that function.  Using default parameters isn't a good programming practice anyways.

MattOttawa:
1000 spots need to be changed. I will do this some time later.

PR #559, opened 2020-04-23
Removed caching of highly used types after loading a game.

MattOttawa:


Thunderbrd:
Just curious but wasn't this one of the most important things we had to reduce turn times?

MattOttawa:
Probably.

Thunderbrd:
So what is the thinking involved in removing them?

MattOttawa:
Removed caching of highly used types after loading a game. #559

These value come from the xml. Caching them once at the start is all that's needed. Caching them again won't give different values, so it's pointless.

Thunderbrd:
I know it seems that way, but from what I understand, it's very slow to access them.  Could be I'm wrong but it's what I've read on the forums.  Alberts2 or Bill would be able to comment further.

alberts2:
Things like the replacement system make it possible to change anything after loading the xml. That makes it necessary to update cached stuff.

PR #572, opened 2020-04-28
Fixed mistake I made while removing BuildingClass

MattOttawa:
I think I need to add a part to set arrays to NULL if pClassInfo's array is NULL and the Info running the func is not and bForceOverwrite is true.


alberts2:
Looks These changes should 

> I think I need to add a part to set arrays to NULL if pClassInfo's array is NULL and the Info running the func is not and bForceOverwrite is true.

That would give some additional memory savings.

alberts2:
@MattOttawa @Toffer90 
This should now take care of the problems in CvBuildingInfo::copyNonDefaultsReadPass2 with the `bForceOverwrite` tag.

Thunderbrd:
Well done!  Thank you Alberts2!

PR #578, opened 2020-05-06
Moving #includes out of CvGameCoreDll.h

MattOttawa:


alberts2:
This should be done the other way around by removing all Cv* and Cy* headers from the precompiled headers. This would really speed up the compilation after making changes.

MattOttawa:
1>1> Obj: **Isolated** C:\Users\uhm\Documents\GitHub\Caveman2Cosmos\Build\Debug\CvGameTextMgr.obj

This is from my vs output. It seems to skip CvGameTextMgr. Any idea what's it about?

raxo2222:
It doesn't skip anything, just does this file separately from what I heard.

Toffer90:
> 1>1> Obj: **Isolated** C:\Users\uhm\Documents\GitHub\Caveman2Cosmos\Build\Debug\CvGameTextMgr.obj
> 
> This is from my vs output. It seems to skip CvGameTextMgr. Any idea what's it about?

Bill said: `Just means that it builds outside of the unity file, i.e. builds the cpp on its own`
When I last inquired about it.

MattOttawa:
I'm actually not sure how I want to do this. Can I add #include "CvGameCoreDLL.h" to most headers and put an include for most files that are not in the pch in one file and share the one file similar to the way CvGameCoreDLL is done? How important is it to keep #includes to a minimum?

alberts2:
You can find some good advice at https://www.acodersjourney.com/top-10-c-header-file-mistakes-and-how-to-fix-them/

MattOttawa:
"If you can guarantee that your code will always be compiled with compiler which understand #pragma once,you use it, otherwise you use include guards."
The guidance is for public header files/ APIs that are distributed to external parties - we don't know if their compiler supports #pragma once or not - so the include guard is necessary.
If you know your partner's compiler supports #pragma once, putting an header guard is not advisable like you mentioned.



You're right Oleksandr - putting in a check like that when the compiler supports #pragma once is a waste.

PR #593, opened 2020-06-25
Parallel maps update

MattOttawa:


MattOttawa:
Anyone wanna check this?
Just go in to assets -> config -> init.xml and comment in the module near the bottom.
Start game and end 1 turn (I think?).
Alt 1 to switch.
Alt 0 to switch back.

This is all I got for this project. It's too hard.

MattOttawa:
This is not saved game compatible I think. I'll modify so it is before I merge.

PR #609, opened 2020-07-08
2 things for nomadic start

MattOttawa:
players can research without any cities and founding cities require tech


MattOttawa:
This is disabled. No change to dll.

PR #613, opened 2020-07-18
irishhombre mod mod.

MattOttawa:
irishhombre: It always irritated me that entertainers could not build all stories so I fixed it. While I was at I also fixed Paleolithic cats and marsupials. Paleolithic cats stories can be built both by cave lions and entertainers. Paleolithic marsupials can be built by both procoptodons and entertainers.


MattOttawa:
This can be reverted.

PR #624, opened 2020-08-11
switching greater gold to long long

MattOttawa:


Toffer90:
I'm pretty sure you need to keep the below line in CvPlayer for it to be save game compatible
`WRAPPER_READ(wrapper, "CvPlayer", &m_iGold);`
but since you're removing the write you must change it to:
```cpp
// @SAVEBREAK DELETE
WRAPPER_SKIP_ELEMENT(wrapper, "CvPlayer", m_iGold, SAVE_VALUE_ANY);
// SAVEBREAK@
```
When something is not written, but must still be read out of old save then the WRAPPER_SKIP_ELEMENT allows that without ending up at the end of the strea prematurely because it didn't find it for saves made after the change.

And also, you should move the read and write of the m_**f**Gold to the bottom of the read write lists for max compatibility I think.

Toffer90:
Thanks for doing this btw, this was on my ToDo list, but that list is never-ending. ^^
Have you tested it in game a bit? Don't have to test it much before merging it, I and you will fix any issues quick if there's any problem.

MattOttawa:
There's more to do. One problem is CvGame::calculateSyncChecksum(). It's a Dll export. I don't know what I'm gonna do about that.
You can check compiler errors to see what's left.

Toffer90:
>One problem is CvGame::calculateSyncChecksum(). It's a Dll export. I don't know what I'm gonna do about that.
Hmm, if the value is small (below e.g. 10 000 000) just int cast it.
If it's bigger then subtract 10 000 000 from it while it is bigger than 10 000 000, the final value which is below 10 000 000 is the one you use.

Toffer90:
lol, looks like the entire calculateSyncChecksum function is dangerously close to int overflowing the way it is set up.
Wonder if we need to rethink it completely, perhaps it is possible to add in more turn slices, more cases to its switch there to limit the size of the checksum result more.

MattOttawa:
What's it for exactly? Just to trigger oos errors? It doesn't seem to be called in the dll or python.

Toffer90:
> What's it for exactly? Just to trigger oos errors? It doesn't seem to be called in the dll or python.

YEah, I think its for triggering OOS errors, though we could do that exclusively in the dll I think. I think we have all the tools necessary to shut the exe out of that functionality.

Thunderbrd:
> What's it for exactly? Just to trigger oos errors? It doesn't seem to be called in the dll or python.

Yeah exactly.  However, it just needs to match to what the other player has is all so Toffer's solution should work fine.

How is this working out when the gold amount needs to be reported to python?

Also, I was thinking you might want to do more commenting out rather than eliminating so that the system can be used again with greater increments once we exceed long long amounts, which we probably will.  The whole concept was to setup a way to then exceed this again with Greater Greater Gold and potentially even Greater Greater Greater Gold assuming in the eventual late game we may really push things.

Thunderbrd:
And I suppose this all means we figured out how to SAVE Long Longs so that's cool.

Toffer90:
> Also, I was thinking you might want to do more commenting out rather than eliminating so that the system can be used again with greater increments once we exceed long long amounts, which we probably will. The whole concept was to setup a way to then exceed this again with Greater Greater Gold and potentially even Greater Greater Greater Gold assuming in the eventual late game we may really push things.

This is the maximum long long, we shouldn't really ever exceed that value in my opinion.
9,223,372,036,854,775,807

An idea I had was to add in a treasury maintenance cost so that the more gold you have in treasury the more you have to pay for keeping that gold safe and secure kinda thing. Will also be an incentive to hurry cost more for players, to invest the money before it waste away.

Toffer90:
> And I suppose this all means we figured out how to SAVE Long Longs so that's cool.

Yeah, Matt suggested that we just convert it to a double and save it as a double, then when the save is read the value is converted back to a long long.

Thunderbrd:
> > Also, I was thinking you might want to do more commenting out rather than eliminating so that the system can be used again with greater increments once we exceed long long amounts, which we probably will. The whole concept was to setup a way to then exceed this again with Greater Greater Gold and potentially even Greater Greater Greater Gold assuming in the eventual late game we may really push things.
> 
> This is the maximum long long, we shouldn't really ever exceed that value in my opinion.
> 9,223,372,036,854,775,807
> 
> An idea I had was to add in a treasury maintenance cost so that the more gold you have in treasury the more you have to pay for keeping that gold safe and secure kinda thing. Will also be an incentive to hurry cost more for players, to invest the money before it waste away.

That's longer than I had been led to believe - perhaps it would be plenty sufficient then.

As for the treasury, that's a great idea and one thing it might then allow is for a building tag that manipulates 'treasury maintenance' expenses so that buildings like the treasury itself might not enhance gold output so much as be used to reduce gold loss from that source.

With a little more application of the inflation mechanism (but fixing it so the penalty properly decays) being put between a rock and hard place with treasury maintenance could be a good thing too.

Thunderbrd:
> > And I suppose this all means we figured out how to SAVE Long Longs so that's cool.
> 
> Yeah, Matt suggested that we just convert it to a double and save it as a double, then when the save is read the value is converted back to a long long. It seems to work just fine, I already did it on the unit upkeep gold amount stored on a national level.

Outstanding - I'd never have figured that out even if I'd been given the suggestion to do it that way.  Well done!

Thunderbrd:
So I guess the next thing would be to convert culture back to a long long instead as well.

Toffer90:
One thing I'm pondering right now is that casting that double back to a long long may yeld different values on different computers as the value is so close to a whole number due to having been cast to a float from an integer in the first place.
Maybe we have to do m_iGold = static_cast<long long>(m_fGold + 0.01) to make sure that different comps round down to the same whole number? We know the float value is clost to a whole number so adding that 0.01 should be enough to ensure that any CPU sees that it is above that whole number and not below it.

Thunderbrd:
> We don't want multiplayer games to OOS the moment the save has been loaded because two comps interpreted the same save file content differently.

Yeah that would be very bad.

MattOttawa:
Maybe bill will look at read / write code we're using for long long.

MattOttawa:
Just gotta do the sync checksum function. Still don't know what to do.

Thunderbrd:
> Just gotta do the sync checksum function. Still don't know what to do.

What needs to be understood with this is that the amount doesn't HAVE to be accurate, just matching what all other systems show and of course based on the actual amount so that it has the relevancy when it's compared to the other systems to detect an OOS.

So just create a loop that manipulates the return on that function so that if it's over MAX_INT, it get's reduced by 1 million until the value is under MAX_INT and then use that result to report as the return.  So long as all systems reporting report the same thing, no OOS should be detected.

Toffer90:
>> Just gotta do the sync checksum function. Still don't know what to do.
> Hmm, if the value is small (below e.g. 1 000 000) just int cast it.
> If it's bigger than that: subtract 1 000 000 from it **while** it is bigger than 1 000 000 (while loop), the final value which is below 1 000 000 is the one you use.
> 
> That should be good enough for the sync checksum.
> Or perhaps better:
> iMultiplier += GET_PLAYER((PlayerTypes)iI).getGold() % 1 000 000;
> 
> I'm thinking that when it goes out of sync the difference in the sync on that value is probably less than a million anyway.

Thunderbrd:
> > > Just gotta do the sync checksum function. Still don't know what to do.
> > > Hmm, if the value is small (below e.g. 1 000 000) just int cast it.
> > > If it's bigger than that: subtract 1 000 000 from it **while** it is bigger than 1 000 000 (while loop), the final value which is below 1 000 000 is the one you use.
> > 
> > 
> > That should be good enough for the sync checksum.
> > Or perhaps better:
> > iMultiplier += GET_PLAYER((PlayerTypes)iI).getGold() % 1 000 000;
> > I'm thinking that when it goes out of sync the difference in the sync on that value is probably less than a million anyway.

I didn't want to suggest something that MIGHT give an odd difference in the rounding between platforms.  The last thing you want to do is CAUSE an OOS by trying to avoid one.  I'm not saying I KNOW that's what would happen here... more that I DON'T know.

Toffer90:
> I didn't want to suggest something that MIGHT give an odd difference in the rounding between platforms. The last thing you want to do is CAUSE an OOS by trying to avoid one. I'm not saying I KNOW that's what would happen here... more that I DON'T know.

There is no rounding in that suggestion, its an int cast of a long long, both are integers.

Toffer90:
> Added overflow protection to CvGame::calculateSyncChecksum(). 401224e

I see Matt already added in an int overflow protection that works just fine, it may still long long overflow when there's so many values added together and multiplied with each other, so I still think the gold should be limited somehow before being used to accumulate to the multiplier value there. But meh, it's a lot better like this than what it was anyway.

Thunderbrd:
I suppose I'm not familiar with % as an operator

Toffer90:
> I suppose I'm not familiar with % as an operator

Modulus operator
[How it Works](https://blog.mattclemente.com/2019/07/12/modulus-operator-modulo-operation.html#how-it-works)
[A typical use](https://blog.mattclemente.com/2019/07/12/modulus-operator-modulo-operation.html#restrict-number-to-range)

Thunderbrd:
The link made it make sense to me - it finds the remainder after a division.  Interesting and surprising how many uses there would be to that.  I'd not heard of this operation before so it's good to know.  I would've been looking to avoid anything that uses a division for this but it should be safe as long as it's not dividing a float and it's not.

MattOttawa:
Looks all good to me now.

PR #627, opened 2020-08-15
revert broken change

MattOttawa:


Toffer90:
Why are you still working on a fork?
Fork is only interesting for people who doesn't have write privilege on the root project, or for those who want to develop the fork in a different direction than the root is being developed in.

MattOttawa:
Hmm.. this aint workin right.

PR #628, opened 2020-08-16
Switched unit get functions so that commanders don't call function again

MattOttawa:


Thunderbrd:
While I don't see a problem here, I'm curious what the effective differences is and why to do one over another.  Also, why are we expanding the restrictive use of const rather than trying to eliminate it?

MattOttawa:
I mod C2C like I play it - on slowest setting. Each step doesn't matter much.

Thunderbrd:
OK, I was just looking for a more technical understanding of the reasoning.  Seeking insight.  That's all.

Toffer90:
> While I don't see a problem here, I'm curious what the effective differences is and why to do one over another. Also, why are we expanding the restrictive use of const rather than trying to eliminate it?

There is little to no difference between calling an instance variable rather than going through a get function.
The get function is mostly there so that the variable can be private and thus not possible to change from the outside without going through a set/change function. Since he's changing the instance to a const it isn't possible to change it's variables, so in that sense there's no need to protect the value in that case.

Using const has the advantage of letting the compiler optimize the code more as it knows that this one or that one will never be changed and thus can apply very specific compilation to it and the surrounding code inside a given scope, if something is not a const when it could be, the compiler still has to compile so that it keeps room for the variable to change even though it won't. That's as good a trivial explanation I can give I think.

Thunderbrd:
I see.  Koshling had basically said there was no advantage so I was just curious.  Makes sense.

MattOttawa:
if someone adds inline to the get function there would be no difference. Seems easy to me, why not inline all get functions? But nobody seems to care about that, so I havn't bothered adding inline much. Maybe the compiler does a good enough job on it's own.

Thunderbrd:
While I have NO idea what an inline is vs a non-inline, I did run across some conversation online a while back that suggested that modern compiling takes care of optimizing for inline vs non-inline whether you state it explicitly or not now.

PR #630, opened 2020-08-16
switch long long back to int for CyPlayer::changeGold to fix bug reported by jatkinson

MattOttawa:


Toffer90:
This is the wrong way to fix the bug reported by jatkinson.
Python is giving a float, so long long and int are both equally wrong in that sense.
The bug is rather in the python code that delivers a float when a long long is expected.

Toffer90:
I fixed a couple changeGold spots in python that delivered float a couple days ago in commit 8b42e3a1226ea20cc3cb8626c22bda354ec5368f

MattOttawa:
You sure you want that a long long anyways? int should be big enough.

Toffer90:
The set function should be capable of setting any valid value.
And looking into it, the bug jatkinson had is already fixed on git, it is one of the spots I fixed.

PR #373, opened 2020-01-12
[WIP] Integer types 1

MattOttawa:
Switching ints to short or char.
Work in progress.

PR #631, opened 2020-08-17
added uint64_t typedef

MattOttawa:


Toffer90:
Nice, it is so annoying to write "unsigned long long".

Thunderbrd:
uint64_t is an improvement on writing "unsigned long long"?  I guess for me I find it vaguely termed and the _ makes it harder to type out. (you have to use a shift)

Anyhow, is this interchangeable with unsigned long long?

Toffer90:
It's still easier and looks better in the code.
@MattOttawa could we perhaps just rename them to int64 and uint64, redact the `_t`?

Thunderbrd:
I suppose as it's one word.

Again though, would it be interchangeable?

MattOttawa:
The _t can be dropped for sure, but it works with the colors for notepad++ the way it is, which is why I like those.

Toffer90:
> The _t can be dropped for sure, but it works with the colours for notepad++ the way it is, which is why I like those.

That's a valid point, I don't want to lose the color coding.

> Again though, would it be interchangeable?

Sure, both variants can be used.

Thunderbrd:
OK, but does one MEAN the other in the code or do they have to be recast to be used in calculations with one another?

MattOttawa:
> uint64_t is an improvement on writing "unsigned long long"? I guess for me I find it vaguely termed and the _ makes it harder to type out. (you have to use a shift)

I still have the branch where I started switching int to short or char. Bill didn't like it, but I havn't forgotten about it. If short and char are used more in the future then I think these typedefs will be nice.
Cause it would be:
int8_t instead of char
int16_t instead of short
int64_t instead of long long

https://github.com/caveman2cosmos/Caveman2Cosmos/pull/373

Thunderbrd:
I can see that being useful though I'm better at remembering the original type names so it would just take a bit to adjust is all - and somewhere I could reference what they mean easily.

Toffer90:
> OK, but does one MEAN the other in the code or do they have to be recast to be used in calculations with one another?

They are equivalent, so one is the same as the other, no need to cast between them when two variables with both are used together. 'long long' casting int64_t would be the same as 'long long' casting a 'long long'., meaning redundant.

PR #632, opened 2020-08-17
food consumed by units

MattOttawa:


Toffer90:
Cool initiative.

Thunderbrd:
Looking at your changes helps me to answer the questions I had about design intent mostly.  Generalizing the expense and then splitting that cost across the empire is probably a good idea.

I would ask that this be done within the frame of a game option.  Not all will want this, no matter how cool we here think it is.  And its easier to set that up now and apply the GC.getGame().isOption(GAMEOPTION_WHATEVAH) if statements as you go.

I will probably need to manage this in a different way with my supply line and nomadic start methods though maybe we can find a way to blend it all together eventually, particularly if the micromanagement of the approach I have in mind there becomes too laborious.  Having the option switch there will help with being able to blend our efforts.

Don't get me wrong, I'm with Toffer and feel this is a great idea - I actually pitched it myself a number of years back almost exactly as you're stepping into implementing it here.  Love it!

MattOttawa:
Units on a city are fed by that city. (Should the owner of the unit be checked?)
The food consumed by units not on a city will be added up then divided by the number of cities that are set to "feed the army".
For that I was hoping Toffer could add a button in the city screen or some way for players to toggle whether or not a city is feeding the army. I'll need a bit of AI for that too...

cbltv1:
I'm assuming this is to reduce city growth throughout the Eras.

Toffer90:
> I'm assuming this is to reduce city growth throughout the Eras.

It would have that effect, but it would hit warmonger players harder than builders.
It will have to be an option, as I'm not sure I would want to play with it much.

Toffer90:
> For Toffer to add such a button, not only will a small graphic be needed, and he can give you an idea of what he wants for that, but you'll need to add a widget. See if you can follow the coding for the widget I made a few years ago for the demolish building/city button.

No widget is needed, all I need is a function exposed to python so that python can tell the dll when the button is pressed.
Perhaps another function to ask the dll whether the button state is active or not so that the button graphic can mirror the state.
On/off state I mean. So, a "**get**" and a "**set**" function or "**is**" and a "**toggle**" function, whatever naming you like best for what it does.

Thunderbrd:
Maybe I don't understand what a widget is for then... my understanding was it was to make it possible to add a UI interface effect.

Toffer90:
> Maybe I don't understand what a widget is for then... my understanding was it was to make it possible to add a UI interface effect.

I always just use WIDGET_GENERAL for all UI, usually I can code out what happens when the UI element is pressed in python.
WIDGETS are only meant to hide all interaction code within the dll, meaning that no python code whatsoever is triggered by the interaction. WIDGET_GENERAL will trigger the handleInput python function within the active screen class when interacted with, while the specific widgets will not inform python of any interaction, just go straight to dll.

In PPIO I replaced a lot of specific widgets with the general one when I saw I could replicate the dll code for its interaction within python.

If it's used for multiple screens, and the code doesn't differ between them, then it can be standardized into a specific widget and have the code centralized in the dll rather than duplicated within multiple handleInput functions for different python screen objects.

Toffer90:
> Is it just to create a hover tooltip perhaps?

It's for that, and for handling what happens when it's clicked.
Though there's no way to get the wide python tooltip without using the GENERAL widget.

Thunderbrd:
It doesn't include sync protection to use the widget does it?

Toffer90:
> It doesn't include sync protection to use the widget does it?

Not sure, I usually write sync protection code in python when interaction changes the game state.
I think most dll widget code does similarly and that the exe does not contain any extra sync protection code integrated in the widget system.

Thunderbrd:
We'll have to discuss someday how you protect changes in python regarding synchronizing.  I've never gotten any insight into that process.  But it's good to know that its a habitual thing for ya.  I suppose if the state wasn't being changed in Python, and we weren't doing it through a widget, we'd still need to setup the state change through the message reporting system.  I guess?

Toffer90:
> We'll have to discuss someday how you protect changes in python regarding synchronizing. I've never gotten any insight into that process. But it's good to know that its a habitual thing for ya. I suppose if the state wasn't being changed in Python, and we weren't doing it through a widget, we'd still need to setup the state change through the message reporting system. I guess?

Python has its own version of:
- `CvMessageControl::getInstance().doSomething(someArguments)`

It's usually used like this in python: 
- `CyMessageControl().sendModNetMessage(905, iPlayer, iCity, -1, UNIT)`

PR #635, opened 2020-08-22
Switched national culture to long long

MattOttawa:


Toffer90:
There's some save compatibility issues, I'm working on them now.

Toffer90:
I think it's ready to be merged now.

Sources/CvPlayer.cpp
```diff
@@ -21933,7 +21905,7 @@ void CvPlayer::read(FDataStreamBase* pStream)
 	// @SAVEBREAK REPLACE
 		double fGold = 0;
 		WRAPPER_READ(wrapper, "CvPlayer", &fGold);
```
MattOttawa, line 21907: don't need these brackets?

Toffer90: what brackets?

MattOttawa: this->m_iGold = static_cast<int64_t>(fGold + 0.01) + (1000000 * m_iGreaterGold) + m_iGold;
		this->m_iGold = static_cast<int64_t>(fGold + 0.01) + 1000000 * m_iGreaterGold + m_iGold;

Toffer90: They are not needed. multiplication always happen before addition.

MattOttawa: ha

Toffer90: i.e. it would be needed if it was like so:
`this->m_iGold = static_cast<int64_t>(fGold + 0.01) * (1000000 + m_iGreaterGold) * m_iGold;`
Because this would not be the same
`this->m_iGold = static_cast<int64_t>(fGold + 0.01) * 1000000 + m_iGreaterGold * m_iGold;`

PR #638, opened 2020-08-25
Switched a couple typdefs.

MattOttawa:
Found this in the code cleanup project:
Standardize the basic types (copy C++11 types?), static_assert sizes.

These integer type defines are part of standard library file <cstdint> (C++11).
File won't work (I guess we're pre c++ 11?)  so I copied them.

Anyways, I plan on switching more types unless someone stops me.


Toffer90:
I won't stop you, word, dword and qword never made much sense to me, better to call int an int imo. ^^
Having the type bit size in the type name is informative and consistent.

PR #641, opened 2020-08-29
fixed trait cost

MattOttawa:


Toffer90:
I see the math is very different in this PR and I guess this is how it was originally, not just a multiplication with 10 anymore, good.

Max level before a long long overflow in the next level threshold value will be much higher with this math, hard to say how much without writing each loop down on paper until it overflows. ^^

PR #646, opened 2020-08-31
switched bForLoad param used for infos

MattOttawa:


Thunderbrd:
Why are you destroying the ability to eliminate this data based on the game options in place?

PR #649, opened 2020-09-03
Parallel maps turn

MattOttawa:


MattOttawa:
I've done a bit of thinking and know how I wanna get started. I don't know if I'll wanna do much more though. It might be too hard. But this may help others figure it out. Next thing to do is go through the players turn and everything called there and move the stuff that needs to run once per map (rather then once per turn) to the new function I added. Hopefully the order of stuff doesn't get shuffled around too much.

Thunderbrd:
Sounds like a good place to focus.  Nice to see progress.  BTW, don't be dissuaded by my initial disgruntlement on the other thread about CvInfos... I just didn't get it at first.

PR #654, opened 2020-09-15
Added Fasserts to FFreeListTrashArray<T>::add and FFreeListTrashArray<T>::getAt

MattOttawa:


Sources/FFreeListTrashArray.h
```diff
@@ -302,6 +302,7 @@ T* FFreeListTrashArray<T>::add()
 	{
 		if ((m_iNumSlots * FLTA_GROWTH_FACTOR) > FLTA_MAX_BUCKETS)
```
MattOttawa, line 303: This is the line that I think acts as a max limit. I think it should be commented out or the max should be raised. Just take out the * 2 out or switch FLTA_MAX_BUCKETS to 20000 or something. See if that fixes SO's crash and push to svn if it does. Then we fix it properly later. My civ4 isn't working and I got no time today. Any1 with GitHub copy of the mod can help.

Toffer90: FLTA_MAX_BUCKETS needs to binary be of the form 10000... only 1 at the beginning.
Raising it from 1 << 13 to 1<< 14 made the save crash while reading the save, so the bit length is hard-boiled in the save it seems. Anyhow, the save still CTD when ending turn after reducing the growth factor.

I'm not sure it's wise to reduce the growth factor without figuring out why its there to begin with.
I could load a save after changing the growth factor from 2 to 1, but I got a lot of asserts from the FFreeListTrashArray object when doing so, most of them from the getAt() function about it not finding anything and returning NULL.

MattOttawa: These FAsserts I added are probly no good.

PR #655, opened 2020-09-18
added null pointer checks after CvPlayer::initUnit calls

MattOttawa:


MattOttawa:
I can pass turn with SO save on this branch. I got a bit lazy and skipped a few spots. I skipped code to merge and split units and never checked python. Partly because I don't really like the fix. If there is going to be a max then it should be checked before the unit is made.

PR #657, opened 2020-09-19
fix(Animal): Give desert tarantula the ability to make same stories as jungle tarantula

dzfischer:
Desert tarantula was missing the ability to create bug - story and bug - stories that jungle tarantula had.


MattOttawa:
This can be reverted if DH doesn't like.

PR #659, opened 2020-09-20
modified beeline stings

MattOttawa:


Thunderbrd:
Well... it's a different way to skin the same cat it seems to me.  So I can't see any reason to object.  Are there any problems this is addressing?

Sadly, this is still something the AI needs a little work to understand how it should shift strategy (prioritize getting to lifestyle techs first until someone gets there and then once someone has it gain strong demotivation to achieve that tech if this option is on.)

MattOttawa:
> Well... it's a different way to skin the same cat it seems to me.

Ya, just that. I was gonna switch the function that part is in to long long and got a little destracted.

Sources/CvTeam.cpp
```diff
@@ -2832,22 +2830,13 @@ int CvTeam::getResearchCost(TechTypes eTech) const
...
-				{
-					iBeelineStingsTechCostModifier += GC.getEraInfo((EraTypes)iI).getBeelineStingsTechCostModifier();
-					//just need to add the tag iBeelineStingsModifier to Era Infos.
-				}
-				else if (iI >= iPlayerEra)
-				{
-					break;
-				}
-			}
+			iCost *= 100 + erasAheadOfTech * GC.getEraInfo((EraTypes)iI).getBeelineStingsTechCostModifier();
```
Toffer90, line 2838: The iI here is not the era index but the player index, which is obviously wrong.
1. You need to loop through all the eras and add each era's beeline sting modifier to the total before using it.
2. Alternatively one multiply the modifier from the players current era by the amount of eras the tech is behind.
3. Alternatively one take the era the tech is from and multiply that with the amount of eras the player is ahead of it.

e.g. if the player is in industrial and the tech is prehistoric.
1. Totaling the beeline sting modifier from each era is then: 100+90+80+70+60+50=450 (this is original design)
2. 5*50=250 
3. 5*100=500
(2 or 3 looks like the design you went for here if I ignore the iI bug where a player index of iI=29 would lead to a NULL eraInfo and error)

PR #661, opened 2020-09-24
Adjustments to lake generation and functionality

dzfischer:
More resources can be spawned on lake tiles in all mapscripts, and additional changes & control for lake generation c2c_world mapscript.
Some cleaning of naval improvement valid terrain types as well.


Toffer90 (approved):
I approve

Issue #418, opened 2020-02-01
Add tools to manage art/audio/text linkage (example: unused files, unused ART/AUDIO/TEXT tags, broken references)


raxo2222:
There may be unused art/audio files - no references to files in mod.
Unused art/audio/text defines - files (TXT_KEY being defined but not used by anything) being referenced, but no actual ingame use.
Items that reference sound/art/text entries, but those entries may be not existing or files not existing.

This tool should consider vanilla files when backtracking to source (thing may be not defined/existing in mod, but be located/defined in vanilla).

TEXT:
Text (TXT_KEY) is defined in text xml files.
`Example: TXT_KEY_BUILDING_ANIMAL_MYTH_REQUIRE is defined in Animals_CIV4GameText and used by zAnimals_CIV4Building(Class)Infos xml.`

ART:
Art (ART_DEF) is defined in artdefines xml files.
They also contain file path entries.
`Example: ART_DEF_BONUS_BARLEY is defined in CIV4ArtDefines_Bonus and used by CIV4BonusInfos.xml
It also references couple files.
<NIF>Art/Terrain/Resources/Barley/Wheat.nif</NIF>
<KFM>Art/Terrain/Resources/Barley/Wheat.kfm</KFM>
<Button>,Art/blank.dds,Art/Interface/Buttons/AtlasMapBonus.dds,1,6</Button>`

AUDIO:
Audio (no consistent tag) are defined in Audio folder.
Sound scripts are used in variety of places, and they use sound defines which then point to files.
`Example:
Some buildings use scriptID AS2D_BUILD_STABLE, which references soundID SND_BUILD_STABLE, which uses <Filename>Sounds/Buildings/Stable</Filename>`

billw2012:
Well we can't verify vanilla files on CI of course, so that would be optional depending on where you run it (unless we just build a database of all valid vanilla values to include).

raxo2222:
It could be simply tool in Tools folder, so you can check if you didn't remove that TXT KEY by accident or if you used art used somewhere else.
First it could search within mod, and then within vanilla assets.

raxo2222:
Seems like I reported it twice...

PR #663, opened 2020-09-27
Organized tech audio xml and files

dzfischer:
OrganizeTechAudio.py can be used in the future to identify mis-linked audio files, or rename the .mp3 files if a different schema is preferred. 
Unused script and define entries were deleted as well as the .mp3 files if there was no tech entry supporting it; some .mp3 files were kept if they had higher quality audio.


MattOttawa:
#418 
Very cool. Too long for me to read though. You gonna expand your python to do more?

dzfischer:
This was mostly written because there's no way I was going to read through 900 * 3 entries to adjust the tech names and see which are linked to nonexistent values, etc. While it is a quick fix type of thing, and not very well optimized or good at catching edge cases, I could likely adapt what's written here to examine other XML linkages if there's one that needs cleaning. If I get into building trees or something, for instance, I'd imagine I might adapt it then but for now I'm going to focus on the lake generation for c2c_world and ofc actually editing the .mp3 tech audio files. Now that they're organized, it'll be far easier 😅

PR #665, opened 2020-09-28
sandball changes: reducing snowball effect

dzfischer:
Tech diffusion: Magnitude is significantly increased!

WFL is unchanged at between +0% and +100% magnitude; new TD is +0% to +100% as well, with better weighting. Welfare (hard era-linked) diffusion unchanged; magnitude 27%*score ratio in prehistoric/ancient dropping to 2%*ratio by atomic/information.

This is a draft PR mostly for an illustrative purpose, to point out how insignificant TD currently is without welfare kicking in as compared to WFL (the actual values proposed here are balanced toward using it directly as an anti-snowball mechanic ala sandball).

image


dzfischer:
The image in the above comment is a comparison of the two TD formulas as they should be implemented; but I'm not a primary c++ coder, so it's entirely possible there is a logic error. "Tech MET" column is the number of civs a player has met that own the technology in question; met war is a subset of those you're at war with, and met open is a different subset of tech met that you have open borders with. These numbers may seem large, but base tech values vary by as much as +- 100% in prehistoric era based on animals subdued, and later can vary by similar amounts based on traits and number/quality of cities.

Thunderbrd:
Interesting.  Will have to look over it eventually but I'm hoping Toffer can take a look first.

dzfischer:
This function currently has unintended behavior at low civ numbers; need to consider how to weight # of teams with tech vs % of world with tech. Does it make sense for TD to be high when there's only one other civ in the world, or low because despite size it is just one civ? Cross purposes.

SannitaSSJ:
Just dropping a line to notice you that I have already imported almost all of the corrections to the text files into /spelling-corrections.

flabbert:
what is the status here? this feels like something that is not resolved still? Should it maybe be turned into an Issue we can track?

dzfischer:
> what is the status here? this feels like something that is not resolved still? Should it maybe be turned into an Issue we can track?

Currently, Tech Diffusion does virtually nothing. This PR was made to give it more teeth for people who wanted it as an option that actually does anything, but unfortunately the impact of Myths on early game is of far, far greater magnitude toward snowballing, to the point that by the time TD comes into play it would need ridiculous percentage boosts in order for it to be a valid catchup mechanic.

Overhauling myths was attempted thru #722 / #772 , but didn't quite integrate properly before moving on, so this got left a bit in the wayside. It will still be necessary to do once myths are overhauled, but the values in this PR will need to be adjusted to whatever the new balance is.

PR #670, opened 2020-10-05
switched plot()->getX() to getX()

MattOttawa:


Toffer90:
669 & 670 both look good to go.

Issue #286, opened 2019-11-29
Planes selected as fort defenders against criminals


billw2012:
https://forums.civfanatics.com/threads/single-player-bugs-and-crashes-v40-plus-svn-after-oct-2019.650924/page-11#post-15606026

dzfischer:
Naval units may also selected as defenders

dzfischer:
Issue still in game, not related to taunt chance; is some issue with coexist rule block and isBlendIntoCity()

dzfischer:
Duplicate and/or subset of #112

PR #676, opened 2020-10-07
Taunt chance no longer takes priority over strength when evaluating 'best defender' in 'target first' combat

### LINKED ISSUE #238: Inconsistent behavior for the armed guard promotion
- ambushers and stalkers ignore armed guard promotions in test case
- assassin ignores armed guard 1, but guard with promos 1-5 selected ; might be visibility issue?
- save: [https://forums.civfanatics.com/threads/single-player-bugs-and-crashes-v40-plus-svn-after-oct-2019.650924/page-6#post-15593027](https://forums.civfanatics.com/threads/single-player-bugs-and-crashes-v40-plus-svn-after-oct-2019.650924/page-6#post-15593027)

dzfischer:
Also added change for hunters to target armed guard units first, following discord discussion.

Linked issue: Resolves #238 
(Also possibly #286 but haven't confirmed)

### DIFF
--- a/Assets/XML/Units/CIV4UnitInfos.xml
+++ b/Assets/XML/Units/CIV4UnitInfos.xml
@@ -5456,6 +5456,10 @@ Base XP gained when defending
 					<UnitCombatTargetType>UNITCOMBAT_ANIMAL</UnitCombatTargetType>
 					<bUnitCombatTarget>1</bUnitCombatTarget>
 				</UnitCombatTarget>
+				<UnitCombatTarget>
+					<UnitCombatTargetType>UNITCOMBAT_PROMOTED_ARMED_GUARD</UnitCombatTargetType>
+					<bUnitCombatTarget>1</bUnitCombatTarget>
+				</UnitCombatTarget>
 			</UnitCombatTargets>
 			<FormationType>FORMATION_TYPE_DEFAULT</FormationType>
 		</UnitInfo>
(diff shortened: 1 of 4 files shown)


dzfischer:
Fixes targeting issue as seen here:
image
Now, armed guard unit properly defends stack against stalker:
image

Previously, getDefenderScore in CvPlot looked at defenderValue in CvUnit, the latter of which gave a +10,000 weight to units that match the 'target first' or 'defend first' criteria; a 10 effective str unit corresponds to ~1000 weight.
However, getDefenderScore _then_ applied the weight from taunt chance. Thus, canine units with inherent taunt would be evaluated as a weight of ~12000 (corresponding to 10,000 from 'target', a handful from combat strength, and mostly from 10-20% added taunt of 10,000). A unit with much higher effective combat strength (say, archer with armed guard) would be evaluated at ~11,000, and so the game engine would choose the dog to defend, despite having often 0% chance of winning compared to archer's 99%.

Two changes were made; first, the taunt weight was moved to defenderValue, and the target override weight applied as the very final item, thus ensuring no other logic acts upon it. Additionally, the target override weight value was increased to 1,000,000 from 10,000; a unit with 300 str and 400% combat odds corresponds to approximately 100,000 defender weight, so the higher weight ensures large combat strength doesn't end up overwriting the ability of target-first to select the appropriate unit set.

It should be noted this is not a good solution (using an arbitrary high value within int range), but passable for now.

MattOttawa:
Looks like you put some effort into figuring this out.

PR #685, opened 2020-10-10
Lake tweaks

dzfischer:
XML tag added; bFreshWaterTerrain, but needs linking to isFreshWater().
Something like
`return algo::any_of(rect(getX(), getY(), 1, 1), GC.getTerrainInfo(getTerrainType()).isFreshWaterTerrain());`
should work, but the syntax is slightly wrong and can't figure out the type requirement algo::any_of is asking for 😅 

Also included (accidentally 🤦) is a hackish fix for the harbor bug; nuked lakes spawning if immediately merging, since the edge cases to check/update are actually somewhat annoying.

### DIFF
--- a/Assets/Python/BUG/PlayerUtil.py
+++ b/Assets/Python/BUG/PlayerUtil.py
@@ -28,9 +28,6 @@
 ##   getPlayerUnits(playerOrID, testFunc), getPlayerCities(playerOrID, testFunc)
 ##     Returns a list of a player's units or cities.
 ##
-##   isSaltWaterPort(city, eAskingTeam)
-##     Returns True if (asking team knows) the CyCity has an adjacent saltwater plot.
-##
 ##   getStateReligion(playerOrID)
 ##   getFavoriteCivic(playerOrID)
 ##   getWorstEnemy(playerOrID, askingPlayerOrID)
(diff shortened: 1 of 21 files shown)


MattOttawa:
I switched loop. I will commit if you find the option that allows people to modify branch.

dzfischer:
image
Seems like you should be able to?

MattOttawa:
Oh ya. Too much pot tonight maybe. IDK if that's what you want exactly.

dzfischer:
image
Ay, works! This is a dual map, and fresh water is being provided to adjacent tile. There's still some more tweaks I'll need to do tomorrow, but thanks Matt for figuring that out for me. Are the foreach_ alterations you've been doing everywhere for performance?
image
Will see about altering the code that does the Fresh Water / Fresh Water Lake effect and tooltip.

MattOttawa:
Not for performance. At first I was just copying bill, but then I kept going for fun.
There is a small bit about loops in the cpp style guide. 
https://github.com/caveman2cosmos/Caveman2Cosmos/wiki/CPP-Style-Guide

dzfischer:
"Do NOT use the legacy code as a guide for good (or even acceptable) style. The legacy code is terrible."
Right... sigh. Not a good place to be editing without a strong sense of what I'm doing >.>

Also lol 'Prefer using algorithms over loops.' Likely not important in this case, but I wouldn't be the one to know for sure.

MattOttawa:
Bill is kinda funny.
I see why he likes those algorithms. I like them too. The way your function is setup it wasn't possible to add it to the algo. There are lots of spots like that in the code like that and making it possible is on my to do list. (just for fun)
For now I wouldn't want any of my stuff gettin in your way.

dzfischer:
image
- Small saltwater 'lakes' no longer provide fresh water, but are still considered lakes for the purposes of events, AI, mapgen scripts, etc
image
Large freshwater lakes (over mapsize definition for lake max) still provide freshwater, but are considered as oceans/coast for the purposes of events, AI, etc.

dzfischer:
There's a few places where `(GC.getTerrainInfo(getTerrainType()).isFreshWaterTerrain() && isWater())` gets reused (in place of `isLake()`), so possibly could be its own function to reduce hardcoding or smth, not sure best practice or how preferred.

dzfischer:
image
Saltwater lakes probably won't appear on current mapscripts, though will probably appear on games from before the lakesize was changed to a function of mapsize. Mostly here to provide a smidge more info the the human that the AI also has access to, plus potential debugging assistance if/when oddities pop up. This is probably it, until Toffer reviews and sees what should be altered/what else should be done.

PR #689, opened 2020-10-14
revert a change I made.

MattOttawa:
Not something any would notice, but not what I mean to do as well.


MattOttawa:
nevermind. this makes sense.

PR #691, opened 2020-10-16
Switching Cy files to return NULL instead of a dummy obj

MattOttawa:


MattOttawa:
Ya, nice. And if I switch all the spots like this to return NULL, then I should be able to take out all the pointer checks in CyCity that every function has I think.

Toffer90:
> Ya, nice. And if I switch all the spots like this to return NULL, then I should be able to take out all the pointer checks in CyCity that every function has I think.

As long as we all agree that there's no reason ever for python to create a null City object on its own with CyCity().
I can't see any reason for why a python modder would want to initialize a CyCity or other such object; I don't think python would be able to tell the dll that this city object now is real and should be on this and that plot.
We have the player.initUnit() and .initCity() for a reason, the dll handles these objects initialization.

PR #693, opened 2020-10-18
Tower spam reduction in city radius

dzfischer:
Towers (but not forts) should not be considered to be built in the workable radius of a city by AI workers. Won't know for sure until a longer test is done, though.

Not sure if there is a scenario for which a tower in workable radius is desirable, but not fort. An extreme edge case may be a 2-tile wide chokepoint, but that'd be hard to check for lol. Visibility range boons may as well be given to units in a city rather than near it.

### DIFF
--- a/Sources/CvCityAI.cpp
+++ b/Sources/CvCityAI.cpp
@@ -12166,8 +12166,8 @@ void CvCityAI::AI_bestPlotBuild(CvPlot* pPlot, int* piBestValue, BuildTypes* peB
 		{
 			bValid = true;
 		}
-		//	Don't count forts - they have their own separate decision criteria
-		else if (!GC.getImprovementInfo(eImprovement).isActsAsCity())
+		//	Don't count forts or towers - they have their own separate decision criteria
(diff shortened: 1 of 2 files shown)


MattOttawa:
cool. I wonder if you should check if the plot is owned by the owner of the unit right after pPlot->isCityRadius()?

dzfischer:
The scenario I'm talking about is when the plot is in the radius of a city (so passes pPlot->isCityRadius()), but that radius happens to be of a civ that's not of the owner. Ex:
C = city owned by player
R = tile in workable radius of city
U = tile not in workable radius of city (at this time)
X = tile in question
O = other city
C-R-R-U-X-R-O
If X is owned by player, but in radius of city O, then this check would fail when perhaps it should succeed. Whether or not the tile is owned by player shouldn't actually matter for this logic (I assume elsewhere workers won't try to improve tiles owned by other civs).

Toffer90:
Could make a new team function, or maybe pass the workers team index in `plot->isCityRadius(eTeam)` and check city owner there, if eTeam is not NO_TEAM.

dzfischer:
Quick test seems to show that this successfully resolves issue with tower spam in workable radius. I'm not confident in the details of how to implement Toffer's suggestion, so I'll just merge this unless objections?
image

MattOttawa:
No objections from me.

PR #695, opened 2020-10-19
switched while loops to for loops for cities in python

MattOttawa:
The function I added to CyPlayer is interesting but might be slower. Gonna have to check.

### DIFF
--- a/Assets/Python/CvAdvisorUtils.py
+++ b/Assets/Python/CvAdvisorUtils.py
@@ -176,8 +176,7 @@ def endTurnFeats(iPlayer):
 				popupInfo.addPopup(iPlayer)
 	# Trade Route
 	if not CyPlayer.isFeatAccomplished(FeatTypes.FEAT_TRADE_ROUTE):
-		CyCityX, i = CyPlayer.firstCity(False)
-		while CyCityX:
+		for CyCityX in CyPlayer.cities():
 			if not CyCityX.isCapital():
 				if CyCityX.isConnectedToCapital(iPlayer):
 					CyPlayer.setFeatAccomplished(FeatTypes.FEAT_TRADE_ROUTE, True)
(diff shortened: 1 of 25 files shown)


Toffer90:
> Gonna have to check.
I can compare them thoroughly tomorrow if you want.

MattOttawa:
That would be good. Problem is there is no resize() method for the list and it probly acts like a vector when you add to it. Unless extend is when I'm lookin for?
https://www.boost.org/doc/libs/1_39_0/libs/python/doc/v2/list.html#classes

Toffer90:
New method is more than 100% faster
```
14:51:01 DEBUG: Timer - OLD took 255 ms
14:51:02 DEBUG: Timer - NEW took 119 ms
number of cities = 93

14:51:02 DEBUG: Timer - OLD took 255 ms
14:51:02 DEBUG: Timer - NEW took 116 ms
number of cities = 93

14:51:04 DEBUG: Timer - OLD took 252 ms
14:51:04 DEBUG: Timer - NEW took 113 ms
number of cities = 93

14:51:04 DEBUG: Timer - OLD took 255 ms
14:51:04 DEBUG: Timer - NEW took 113 ms
number of cities = 93

14:51:05 DEBUG: Timer - OLD took 255 ms
14:51:05 DEBUG: Timer - NEW took 113 ms
number of cities = 93

14:51:05 DEBUG: Timer - OLD took 256 ms
14:51:06 DEBUG: Timer - NEW took 120 ms
```

Toffer90:
For any python modders that are unsure about how to compare performance of two methods in a proper way:
```py
CyPlayer = GC.getActivePlayer()
print "number of cities = %d" % len(CyPlayer.cities())
timer = BugUtil.Timer('OLD')
for i in xrange(6000):
	CyCityX, i = CyPlayer.firstCity(False)
	while CyCityX:
		a = CyCityX.getOwner()
		CyCityX, i = CyPlayer.nextCity(i, False)
timer.log()
timer = BugUtil.Timer('NEW')
for i in xrange(6000):
	for CyCityX in CyPlayer.cities():
		a = CyCityX.getOwner()
timer.log()
```

dzfischer:
`timer.log()` -- what file does that actually log to, is my question? 😄

Toffer90:
> `timer.log()` -- what file does that actually log to, is my question? 😄
> Was looking for it earlier, but couldn't seem to find.

PythonDbg.log
To get the test to repeat infinitely without anything other than the test being written to that log file you just need to put the comparison code within CvMainInterface.updateScreen(); a function that is continously called in the background every 250ms when you're at the normal map view. There are other good places to put it, but that's a pretty base place to do so at.

PR #703, opened 2020-10-23
added a break; in CvCity::doAutobuild()

MattOttawa:
break; should help, but probably not the main problem.
Would be simpler for a dummy like me if the same evaluation was used for adding and removing auto buildings.
Evaluate if building should be free for the city first, then check if city has building and if a change is needed.

Maybe I'm over simplifying.


Toffer90:
The break is an improvement, but as you said, not the main problem, would just cause one or more duplicate messages in rare cases (very few autobuilds require a specific amount of more than one building type).

Could you correct the comment typo `Speciasl → Special` while you're making changes in there. 
Thanks.

PR #708, opened 2020-10-25
cleaned up m_paiBuildProgress read/write (breaks saves)

MattOttawa:


Sources/CvPlot.cpp
```diff
@@ -11839,7 +11771,7 @@ void CvPlot::write(FDataStreamBase* pStream)
 	}
 	else
 	{
-		WRAPPER_WRITE_DECORATED(wrapper, "CvPlot", (char)MAX_PLAYERS, "cFoundValuesPresent");
+		WRAPPER_WRITE_DECORATED(wrapper, "CvPlot", (char)1, "cFoundValuesPresent");
```
Toffer90, line 11774: This will bug out on the other end without further changes:
```cpp
	WRAPPER_READ(wrapper, "CvPlot", &cFoundValuesPresent);
	if ( cFoundValuesPresent != (char)-1 )
	{
		unsigned int* m_aiFoundValue = new unsigned int[cFoundValuesPresent];
		WRAPPER_READ_ARRAY(wrapper, "CvPlot", cFoundValuesPresent, m_aiFoundValue);
	}
	else
```

Toffer90:
This PR cannot be accepted at this time as it breaks saves.

Sources/CvPlot.cpp
MattOttawa: why?

Toffer90: Because you changed cFoundValuesPresent to equal 1 instead of MAX_PLAYERS.
m_aiFoundValue = new unsigned int[1] is not the same as m_aiFoundValue = new unsigned int[MAX_PLAYERS]

MattOttawa: wasn't paying that much attention.

PR #710, opened 2020-10-25
SaveBreak pre v41 release

Toffer90:
To be merged when we agree we're close to a v41 release.

Anyone is welcome to expand on this branch before it's merged.


raxo2222:
Someone could rename all buildings and units to match ingame name.

dzfischer:
> Someone could rename all buildings and units to match ingame name.
Would be nice... I don't suppose there's a tool anywhere that does this generically? I could try to adapt the script I wrote that does this for techs to general xml usage, but might take a bit.

raxo2222:
> 
> 
> > Someone could rename all buildings and units to match ingame name.
> 
> Would be nice... I don't suppose there's a tool anywhere that does this generically? I could try to adapt the script I wrote that does this for techs to general xml usage, but might take a bit.

There is no such tool.
Essentially it would have to collect displayed names, and notice those in each unit or building entry.
``
<UnitInfo>
			<Type>UNIT_X</Type>
			<Description>TXT_KEY_UNIT_XX</Description>
....
</unitinfo>
``
Then it could print nice 10 000 entry long CSV with ingname name, description tag, and type tag.

PR #713, opened 2020-10-29
Misc tweaks

dzfischer:
Scrap cost currently doesn't seem to work when grouping down, due to odd behaviour with pow function; pow(3, -1) returns 0 for some reason?

### DIFF
--- a/Sources/CvUnit.cpp
+++ b/Sources/CvUnit.cpp
@@ -7026,6 +7026,26 @@ bool CvUnit::canScrap() const
 }
 
 
+int CvUnit::calculateScrapValue() const
+{
+	float fCost = float(getUnitInfo().getProductionCost() * GC.getGameSpeedInfo(GC.getGame().getGameSpeedType()).getTrainPercent() / 100);
+	fCost += fCost * float(getUpgradeDiscount() / 100);
+	logBBAI(" fCost is %f", fCost);
+
+	if (GC.getGame().isOption(GAMEOPTION_SIZE_MATTERS))
+	{
+		int iGroupDiff = groupRank() - m_pUnitInfo->getBaseGroupRank();
+		logBBAI(" iGroupDiff is %i", iGroupDiff);
+		double dSizeMod = std::pow(3, iGroupDiff);
+		logBBAI(" dSizeMod is %f", dSizeMod);
+		fCost *= float(dSizeMod);
+		logBBAI(" fCost is now %f", fCost);
+	}
+
+	fCost /= GC.getUNIT_GOLD_DISBAND_DIVISOR();
+	return std::max(1, int(fCost));
+}
+
 void CvUnit::scrap()
 {
 	if (!canScrap())
@@ -7042,15 +7062,7 @@ void CvUnit::scrap()
 
 	if (GC.getGame().isOption(GAMEOPTION_DOWNSIZING_IS_PROFITABLE) && plot()->getOwner() == getOwner())
 	{
-		int iCost = (getUnitInfo().getProductionCost() * GC.getGameSpeedInfo(GC.getGame().getGameSpeedType()).getTrainPercent()) / 100;
-		iCost /= GC.getUNIT_GOLD_DISBAND_DIVISOR();
-		iCost += (iCost * getUpgradeDiscount())/100;
-		if (GC.getGame().isOption(GAMEOPTION_SIZE_MATTERS))
-		{
-			iCost = applySMRank(iCost, getSizeMattersOffsetValue(), GC.getSIZE_MATTERS_MOST_MULTIPLIER());
-		}
-		iCost = std::max(1, iCost);
-		GET_PLAYER(getOwner()).changeGold(iCost);
+		GET_PLAYER(getOwner()).changeGold(calculateScrapValue());
 	}
 
 	getGroup()->AI_setMissionAI(MISSIONAI_DELIBERATE_KILL, NULL, NULL);
--- a/Assets/XML/Units/CIV4SpawnInfos.xml
+++ b/Assets/XML/Units/CIV4SpawnInfos.xml
@@ -312,7 +312,6 @@ Using both <iGlobalTurns> and <iTurns> will not yield predictable results:
 			<TerrainTypes>
 				<TerrainType>TERRAIN_MARSH</TerrainType>
 				<TerrainType>TERRAIN_MUDDY</TerrainType>
-				<TerrainType>TERRAIN_SALT_FLATS</TerrainType>
 			</TerrainTypes>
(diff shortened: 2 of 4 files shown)


dzfischer:
Maybe Neanderthal Culture wonder should get the corresponding Local Culture free in the founding city?

dzfischer:
Also, are the Adopted Cultures counted as national wonders due to <iMaxPlayerInstances>1</iMaxPlayerInstances>? That probably shouldn't be needed, as they also require Palace, and if they are counted as Nat'l wonders that seems to have minor drawbacks/oddities for limited wonders gameoption as well as for traits that give benefits for national wonders.

Marking Toffer as reviewer request b/c of the Culture changes, maybe you have some other plans in mind or something. If it looks good to you, go ahead and merge, I'm done with these changes (though I'll probably pop open this branch/pr now and then as I find other misc oddities while playing).

Toffer90:
> Scrap cost currently doesn't seem to work when grouping down, due to odd behaviour with pow function; pow(3, -1) returns 0 for some reason?

Probably because the result is rounded down from 0.333 to 0 due to being stored in a int variable or something.

dzfischer:
> > Scrap cost currently doesn't seem to work when grouping down, due to odd behaviour with pow function; pow(3, -1) returns 0 for some reason?
> 
> Probably because the result is rounded down from 0.333 to 0 due to being stored in a int variable or something.

Bill just confirmed on discord it's literally a bug in the math.h implementation used lol

Sources/CvUnit.cpp
```diff
@@ -7026,6 +7026,30 @@ bool CvUnit::canScrap() const
...
+		// Workaround because pow() seems to return 0 when having a negative exponent, for some reason??
+		// Otherwise just "fCost *= float(std::pow(3, iGroupDiff));" would work by itself
+		if (iGroupDiff >= 0)
+		{
+			fCost *= float(std::pow(3, iGroupDiff));
+		}
+		else
+		{
+			fCost *= float(1 / std::pow(3, -iGroupDiff));
+		}
```
Toffer90, line 7046: pow(int, int) returns a int doesn't it?
try changing it to:
 fCost *= 1 / std::pow(float(3), -iGroupDiff));

Sources/CvUnit.cpp
```diff
@@ -7026,6 +7026,30 @@ bool CvUnit::canScrap() const
 }
 
 
+int CvUnit::calculateScrapValue() const
+{
+	float fCost = float(getUnitInfo().getProductionCost() * GC.getGameSpeedInfo(GC.getGame().getGameSpeedType()).getTrainPercent() / 100);
+	fCost += fCost * float(getUpgradeDiscount() / 100);
```
Toffer90, line 7032: The float cast here should be done on only the dividend or the divisor if you want a precision result from that division math. else it will be float(int / int) = float(int) which is pointless as the rounding down happens before you float cast it and because it will automatically make it a float as it is to be stored in a float variable there.

dzfischer: Mmm, right. Will make this change and see if I made other odd choices of the same vein elsewhere, plus altering the pow() func. to how Bill suggested.

Toffer90:
> Also, are the Adopted Cultures counted as national wonders due to `<iMaxPlayerInstances>1</iMaxPlayerInstances>`?

Nope, because they have <bNoLimit>1</bNoLimit>, which means they don't count as a national wonder.

Sources/CvPlayer.cpp
```diff
@@ -7633,7 +7633,8 @@ int CvPlayer::getProductionNeeded(UnitTypes eUnit) const
 	iProductionNeeded *= iModifier;
 	iProductionNeeded /= 100;
 
-	if (GC.getGame().isOption(GAMEOPTION_SIZE_MATTERS))
+	// Trade units aren't impacted by SM production reduction due to ROI oddities. They can't merge anyway, so...
+	if (GC.getGame().isOption(GAMEOPTION_SIZE_MATTERS) && GC.getUnitInfo(eUnit).getUnitCombatType() != GC.getInfoTypeForString("UNITCOMBAT_TRADE"))
```
dzfischer, line 7637: Tested this is functional ingame, but not sure if there's a better way to do this as it's hardcoded to "UNITCOMBAT_TRADE" and seems a bit hacky? Otherwise, I think this PR is ready; changes since yesterday are mostly text, and some value edits as per discord discussion.

Sources/CvUnit.cpp
```diff
@@ -7026,6 +7026,22 @@ bool CvUnit::canScrap() const
...
+{
+	float fCost = getUnitInfo().getProductionCost() * GC.getGameSpeedInfo(GC.getGame().getGameSpeedType()).getTrainPercent() / 100.0f;
+	fCost += fCost * getUpgradeDiscount() / 100.0f;
+
+	if (GC.getGame().isOption(GAMEOPTION_SIZE_MATTERS))
+	{
+		int iGroupDiff = groupRank() - m_pUnitInfo->getBaseGroupRank();
+		// pow(int, int) returns int, not double...
+		fCost *= std::pow(3.0f, iGroupDiff);
+	}
```
dzfischer, line 7039: 👍 / 👎 ?

Toffer90: > 👍 / 👎 ?
> I have no internal sense of perf issues lol. Is using pow(double, int) better than pow(float, int) and such, or float x = int / double vs or float x = int / float, etc? Doubt it matters, but don't know 🤷

Try to not ever use any floating point math where game state is concerned as it can cause OOS issues, multiply with 100 before starting and divide by 100 at the end.

Sources/CvPlayer.cpp
Toffer90: > Tested this is functional ingame, but not sure if there's a better way to do this as it's hardcoded to "UNITCOMBAT_TRADE" and seems a bit hacky? Otherwise, I think this PR is ready; changes since yesterday are mostly text, and some value edits as per discord discussion.

Is there not a canMerge() function? would think there is as some units can't merge.

Sources/CvUnit.cpp
Toffer90: made a PR to this PR that resolves the OOS potential of this code.

dzfischer: Oof, that is right. Will look at my other commits to cpp code, may have made same mistake elsewhere...
If you're done fixing all my terrible code (for which I am eternally grateful...) should I merge the PR(s), or are you still looking at things?

Toffer90: Merge away if you have no objections, I'm all done I think.

I don't have write permission, so I can't merge the PR.

PR #717, opened 2020-11-01
added code to read/write long long

MattOttawa:


Toffer90:
Looks very similar to the branch I made around the last save break period.
I found that the exe CTD'ed hard whenever saving when using it, never got to try out loading saves a sI couldn't save one to begin with even.

Tested this change here on the new savebreak branch just now and I got a CTD at the initial autosave of turn 0.
Maybe I'm wrong about the exe thing and there's a different cause to the CTD that you can fix.
Just thought you should know that this doesn't work as it is.

PR #724, opened 2020-11-05
Unit sm counting

dzfischer:
Two issues:
1- Merging/splitting a unit initializes a template first, then assigns promotions, so unit count isn't properly updated. There could be an update when the promos are assigned, but haven't yet figured out exactly at what point it does so, and feels kinda hacky (though the whole merging/splitting system is anyhow, so...)
2- This breaks selecting tech. Somehow handleInput from CvScreensInterface.py (line 475) dies, and don't know why?! From an unidentifiable C++ exception that isn't causing any compile errors.


Toffer90:
You should split changeUnitCount and getUnitCount so that there's changeUnitCountSM and getUnitCountSM alongside those two.
Two different counters. Knowing the amount of actual units currently on map may be useful even for an SM game.

dzfischer:
Right, almost forgot about this, hah. Feel free to take over if you have interest Toffer, don't worry about stepping on my toes or anything; I have a tendency of dropping things as manic phases come and go, so can't make/uphold any promises I'll get to this particular item before v41 release >.>

Toffer90:
I was a bit worried you'd might get a bit offended, seeing as I pretty much completed it rather than nudge you in the right direction; there was so many merge conflicts with the current master (due to the saveBreak branch touching so much of the same code) that I had to do a full rewrite to even get it to compile.

You could look into if there's more spots in the code where the SM counter should be used, and don't worry about time, this won't be merged to master before after v41 release.

PR #728, opened 2020-11-11
Route change fix

MattOttawa:
gotta find a way to test. I'm thinking I'll wait till after next version for this.


Toffer90:
It looks correct, but no way to check without a save where the interstate event has fired.
I agree that it should wait till after v41 release.

MattOttawa:
That will take a while. I close to keep pull request page clean for now.

dzfischer:
Just saw this. If you get around to working on it again, could you consider adding support for an xml entry for tech change to flat movement cost as well? Currently it works like:
```
<TechMovementChanges>
	<TechMovementChange>
		<PrereqTech>TECH_SKYROADS</PrereqTech>
		<iMovementChange>-10</iMovementChange>
	</TechMovementChange>
</TechMovementChanges>
```
but if there could be an iFlatMovementChange entry, that'd make this much more useful. With both, it should be possible to add a bit of granularity to tech improvement to routes, as increasing iMovement without iFlatMovement doesn't really work that well except for some edge cases.

flabbert:
this feels like an issue not a PR

MattOttawa:
I just need to test this. Need a save with interstate event done (which may be hard to find).

PR #732, opened 2020-11-15
FA_MEN_SI fix

MattOttawa:
this is easiest way for me.


Toffer90:
Better to keep these unique building effects in python rather than hardcoding them in dll imo.

MattOttawa:
If nobody feels like switching around the python it's done here.

Toffer90:
> If nobody feels like switching around the python it's done here.

No need to switch around python... is there?
I already fixed the Fa Men Si message bug in this commit:
https://github.com/caveman2cosmos/Caveman2Cosmos/commit/5a2bebc02aab6b6c631c707e8b9af1f745c5df0e#commitcomment-44185889
 - before you created this PR.

PR #738, opened 2020-11-20
team member cache for CvTeam

MattOttawa:
Made this a couple months ago but never finished. Just gotta make code to add IDs to cache.


Toffer90:
Ah, cool, I'll stop doing what I'm doing with the isAliveAndTeam() function so you don't get more conflicts than there already is.

MattOttawa:
Never mind actually. I've always been a bit afraid of skrewing up part where I add and removed players from vector.

PR #747, opened 2020-11-25
added back part I deleted a long time ago to fix ErmanTheTurk

MattOttawa:
doesn't work.


Toffer90:
Cool, glad the error made sense to someone, it confused me.

MattOttawa:
Na, it still crashes with this change.
This is old bug that nobody could fix.

Toffer90:
The crash point does come later in the code with this change though.

MattOttawa:
Ya it's probly debug code that never worked along with CvPlayer line 20219, and corrupted adjustment in FreeList files.
Something broken about save load for selection groups and all this code was added to fix after game is loaded, and it's still broken.

PR #759, opened 2020-12-01
deleted Exposed to Python comments

MattOttawa:
They arn't correct after I removed unused functions from interface.


Toffer90:
They're pretty dumb to have there in any case. ^^

PR #760, opened 2020-12-01
Improvement(+related UI) fixes, balance tweaks, and organization

dzfischer:
Stronger support for FeatureChange tag; it will not remove all improvements on the tile by default (only prevents itself from leaving an empty improvement), slightly better UI tooltip.
Need to add 3d art from beacon/lighthouse improvements to the features, though.
Likely needs a bit more testing.


dzfischer:
Figuring out how to teach the AI to use this is tricky. Unless I'm mistaken, the meat is in AI_bestPlotBuild which in turn compares all improvements to determine which is best for each plot. There is an "aiDiffYields" calculated, so tossing in improvements whose corresponding build has FeatureChange or TerrainChange with a large multiplier ought to teach the AI to use it properly, but... AI_bestPlotBuild is something of a monstrosity, not gonna lie, don't think I'm quite capable of doing it without serious/slow investigation 😓 
The tooltip is easier, just about finished making adjustments so that it reads from the feature change rather than dummy entries in the improvement yield (which are incorrect depending on current plot terrain/features, ofc...).
There's another issue where "Requires Dummy Tech" is visible to the player in certain circumstances due to the Feature Struct implementation (not introduced by these changes); will add that to git issues when I finish with further detail.

Thunderbrd:
It's all fun until it comes down to having to get the AI to work with it.  lol.

dzfischer:
Old organization of worker actions was pretty darn random:
image
New should be much improved:
image

dzfischer:
Juuuust about done.

I managed to break "Requires Irrigation!" UI warning and I don't know how.... after fixing that and a bit more testing, will merge with master tomorrow.

PR #768, opened 2020-12-08
switched array to vector

MattOttawa:
just gotta do xml I think.


MattOttawa:
too much xml

Toffer90:
I can do the xml here later today.

Toffer90:
Hmm, Blaze's recent unit xml file splitting broke the xml work here.
If you revert the xml part I contributed I'll redo the xml convertion.

dzfischer:
Whoops.
Looks like the XML half of things changed is just a simple find-and-replace though, right?

Toffer90:
It is, 
Takes about 2 minutes to redo the xml work.

PR #772, opened 2020-12-11
Myth from Toffer's MToS

dzfischer:
Still **requires**:
- Removing savebreaking things (CIVIC_SUPERSTITION and similar)
- Sanity gameplay check thru prehistoric era

Should do:
- See if can prevent _massive_ tooltip on Oral History / Supersition civic building that fills screen...
- Examine / apply tech cost adjustment if necessary due to different beaker amounts earlygame
- Unify Taxidermy animal content 

Else?

Is heavy balance adjustment... but that's what the SVN ver is for, right? Or should this be held off until/if we release a v41.1?


flabbert:
closing this PR, I have made a milestone related to this

PR #773, opened 2020-12-11
Periodical pedia cleanup

SannitaSSJ:
**Bonus_CIV4GameText.xml**
* merging alt-timeline's bonuses and all other bonuses scattered in other files into main file (does this break anything?);
* adding command for color highlight for bonuses' names (_color hue still TBD_);
* sorting alphabetically all bonuses (again);
* adding/rewriting pedia entries for Atompunk (en, it), Biopunk (en, it), Sabbatean (fr, it);
* fixing the usual typos and other stuff.

**other files**
* fixing/rewriting several poorly translated entries for fr, es, it;
* removing empty/outdated pedia entries (mostly fr, it, pl);
* removing several fake "French" translations (i.e. English entries in a <French></French> tag)


dzfischer:
> merging alt-timeline's bonuses and all other bonuses scattered in other files into main file (does this break anything?);
Pretty sure this shouldn't break things, but I thought the original purpose of the alt-timeline content being separate was that people could disable if if they need. Granted I'm not sure that's possible any more.... I for one would be fine with moving more "core-ish" module content into base to make things way simpler, but don't know teams thoughts?

MattOttawa:
Ya, keeping things in modules so they can be turned off can be a good way for players to save memory / improve perf.
Then again it is simpler for us modders to not have modules imo.
So it's a bit of a toss-up.
Modules should be turn-off-able I think or they pointless?

Thunderbrd:
I think he's talking about text only.  Rax changed it so we have all texts in one folder for easier searching and it isn't all that much gain to ensure that text is all in their perspective module folders unless you wanted to be able to move a module from one mod to the next which C2C is usually a little too unique to make our modules interchangeable with other mods modules without tons of adaptation anyhow.  However, it IS probably at least a courtesy, in case someone WAS trying to make such an adaptation of one of our modules to another mod, that the text files be in their own modular text folder even if within the core XML->Texts folder.  Turning off and on text files themselves has little or no difference alone though.

If we are talking about more than text files than eliminating modularism is usually not cool since it took more work to set things up to BE modular and is supposed to provide the player with the option to turn off that module if all is setup correctly.

SannitaSSJ:
I'm neutral about it. I mean I can easily revert my changes, but then IMHO we should divide the whole Bonus file into at least three separate files: one for Civ bonuses, one for manufactured bonuses, and one for "basic" bonuses. Plus the modules, of course.

dzfischer:
> If we are talking about more than text files than eliminating modularism is usually not cool since it took more work to set things up to BE modular and is supposed to provide the player with the option to turn off that module if all is setup correctly.

I definitely feel like there's a number of modules that are too ingrained into our system to be removed already though. DH's base animal work, Pepper's space content, etc, really ought to be just wholesale moved into core imo. Things like alt timelines are a bit more iffy...

Thunderbrd:
> > If we are talking about more than text files than eliminating modularism is usually not cool since it took more work to set things up to BE modular and is supposed to provide the player with the option to turn off that module if all is setup correctly.
> 
> I definitely feel like there's a number of modules that are too ingrained into our system to be removed already though. DH's base animal work, Pepper's space content, etc, really ought to be just wholesale moved into core imo. Things like alt timelines are a bit more iffy...

I'm not sure I yet consider anything about the space stuff to be baked in enough to want that in the core yet.  For some things, it's maybe easier to have them modular for the sake of it being separated by project expanse.  DH probably did a lot to ensure true modularism to the animal folders but yeah probably a lot of that COULD or probably should be core - except the special rider stuff is a common request to be able to turn off - though perhaps that would be better setup as a game option to activate those things instead of modular.

MattOttawa:
Actually switching folders might mess up overwriting of bts xml...

PR #786, opened 2020-12-18
switched copy non default code for building prereqs

MattOttawa:


Toffer90:
Should probably do this for the vectors above and below too.
getNumPrereqInCityBuildings
getNumReplacementBuilding

Toffer90:
Tested change, fix work.

MattOttawa:
Thanks.

PR #787, opened 2020-12-19
switched copy non default code

MattOttawa:


Toffer90:
Do you consider this PR complete Matt, or are you planning to do more here?

MattOttawa:
Ya. This may cause some in-game changes though.

PR #798, opened 2020-12-29
Revert "switched some copy non defaults code (#796)"

MattOttawa:
This reverts commit fe77686713af11662e3fb2160aceafb898aa3c41.

### DIFF
--- a/Sources/CvInfos.cpp
+++ b/Sources/CvInfos.cpp
@@ -12980,7 +12980,10 @@ m_iAIInflationPercent(0),
 m_iAIWarWearinessPercent(0),
 m_iAIPerEraModifier(0),
 m_iAIAdvancedStartPercent(0),
-m_iRevolutionIndexPercent(0)
+m_iNumGoodies(0),
+m_piGoodies(NULL)
+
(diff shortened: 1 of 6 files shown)


Toffer90:
You sure it's best to revert all of the commit?

MattOttawa:
Only bug in the last version I can remember is the dynamic civ names one. Might be more bugs in next release, especially without freeze.

Toffer90:
> Only bug in the last version I can remember is the dynamic civ names one. Might be more bugs in next release, especially without freeze.

Gameplay changes has for the the most part been withheld since the improvement/routes/worker-build changes blaze did, which was right before we decided to do another freeze period in preparation for a re-release.
v41 had a repeatable and unavoidable game freeze bug in some unit AI code that appeared quite early in the game.
The game also CTD'ed whenever a city was liberated, either by trade, or when capturing a city and deciding to return it to original owner in the conquest pop-up or when hitting the liberate button in the city screen.
There's been several other minor game fixes, too.

v41 was just broken; hopefully we haven't introduced new CTD/freeze bugs between v41 and v41.1.

PR #800, opened 2020-12-30
Added Russian language tags (svn11281) to text XML files

drd-avel:
Need to merge this to Master and then we can continue....


SannitaSSJ:
Uhhh, @drd-avel it might be better if you contribute directly to the spelling-corrections branch. I moved much of the "new files" to other files, and probably it would be better for us to cooperate on the same branch! :)

Toffer90:
> Uhhh, @drd-avel it might be better if you contribute directly to the spelling-corrections branch. I moved much of the "new files" to other files, and probably it would be better for us to cooperate on the same branch! :)

I'll move it over so it is a branch on your spelling branch instead of a branch to master.
Cool if you take lead in reviewing this work.

PR #802, opened 2021-01-01
Discovery odds visible UI

dzfischer:
Not functional atm, couldn't figure out what I was doing wrong with CvDLLWidgetData.cpp, specifically the line (which does need to be tweaked; use game setting depenent rounding value for instance, but): `iRoundedOdds = 1000 + (pMissionPlot->calculateResourceDiscoveryOdds(iI, GC.getImprovementInfo(eImprovement)) / 1000) * 1000;`

Example log:
```
[272565.063] Discovery of bonus possible!
[272565.063] Rounded odds are: 3000
[272565.063] First buffer
[272565.063] Second bit
[272565.063] Final?
[272565.063] Discovery of bonus possible!
[272565.063] Rounded odds are: 8000
[272565.063] First buffer
[272565.063] Second bit
[272565.063] Final?
[272565.438] Discovery of bonus possible!
```
(Crashes before reporting rounded odds on 3rd bonus)

Also would need to sort resources by iRoundedOdds to reduce lines that'd otherwise be significantly repeated/look ugly. Here for after v41.1

For improvement Z having resource discovery for resources A, B, C, D, and E, with:
Resource : base odds : actual odds
A : 1000 : (can't be placed on current tile)
B : 1000 : 1900
C : 1500 : 2950
D : 1500 : 2900
E : 2000 : 1800

In v41, currently shows:
- Small chance of: A, B
- Small chance of: C, D
- Small chance of: E

Behavior with this current PR (bugged atm):
- Each turn when worked, has approximately:
1/2000 chance of: B
1/3000 chance of: C, D
1/2000 chance of: E

Desired behavior:
- Each turn when worked, has approximately:
1/2000 chance of: B, E
1/3000 chance of: C, D


flabbert:
Should this be an issue instead?

PR #806, opened 2021-01-17
CI: Add pylint for python 2 and 3

trym-b:
This commit adds the possibility to run `pylint`-linting on all python
modules. `pylint` can be easily be run through `docker` by running this
on Windows in Git Bash or on Linux:
```
docker run --rm --tty --interactive --volume <path-to-repo>/Caveman2Cosmos/:/host_dir --workdir /host_dir ubuntu:20.04 bash -c "./Tools/Lint/setup.sh && python3 Tools/Lint/pylint.py"
```
It takes about 3 minutes to run the whole setup, from `setup.sh` is
started until linting has been completed.

As of this iteration, all `*.py`-files are passed to pylint, but all
files listed in `Tools/Lint/python2_modules.txt` are treated as
`python 2`, while every other is treated as `python 3` files. So if
someone adds a new *.py module in the future it will automatically be
linted as a python 3 file.

Both `python 3` and `python 2` has two separate pylint configuration
files, since `pylint` since version 1.9.5 has diverged quite a bit.
All triggered errors has been disabled, any errors that has not been
triggerd are enabled.
Each error ignored in the configuration file should be enabled
one-by-one, some might be trivial to fix, others not.

As part of the first iteration, all modulles are linted in one go, even
if no modules are touched. The reason for this are that `pylint` looks
at multiple modules in one go, so if errors are done cross modules they
might not be detected if pylint is run on just a single module. Also, in
case pylint is updated to the newest version, erros will not be
discovered until someone actually modifies that file.

This is the interface for the new script `pylint.py`:
```
python3 Tools/Lint/pylint.py --help
usage: pylint.py [-h] [--python-versions {2,3} [{2,3} ...]]

optional arguments:
  -h, --help            show this help message and exit
  --python-versions {2,3} [{2,3} ...]
                        Lints all modules targeting a specific python version, will lint both
                        versions by default
```

`pylint.py` can be run locally as well, just be aware that it requires
both `python 2` and `python 3` installed, and it might mess up your
system pip.


MattOttawa:
Can you get it to run after each commit the same way appveyor does?

trym-b:
Sure.

I am unable to trigger Appveyor from my fork. Since I am not super familiar with `Appveyor`, I probably need to test run it a few times. I could create a draft PR and do my testing there, but that will probably spam everyone with a bunch of notifications. If that is acceptable then I can try that.

trym-b:
Alternatively, we could use create Github Actions workflow for doing misc tasks.
Just create a workflow with the path: `.github/workflows/commit.yml`, this might require some configuration on Github, but it would make it easier to read the `appveyor.yml` file

Toffer90:
Heisann Trym, lenge siden. Tok litt tid før det gikk opp for meg hvem nykommeren var. Kult å se deg bidra til C2C, har du fulgt med på utviklingen av modden lenge?

Hilsen Kristoffer (avataren min er ikke like beskrivende som din er ^^)

trym-b:
Jobb og slikt gjør at man ikke kan være like anonym som tidligere. ;) Er på serveren ja, sendte deg akkurat en melding.

PR #808, opened 2021-01-17
Enable f-string-without-interpolation error

trym-b:
This PR illustrates how to enable more `pylint` errors. In this particular a trivial (and hopefully uncontroversial) error is enabled.

See the commit messages in the individual commits.

TLDR: One commit enables a pylint error, the other commit fixes the code


trym-b:
You might want to enable Github Actions workflows to be run on incomming PRs from outside the repo, otherwise you will not know if the PR passes the now brand new linting workflow. Unless there are some other concerns that I am unaware of.

MattOttawa:
I don't know how to do that. Wanna get that done for us please?
Maybe Toffer can add you to github team if that is needed.

trym-b:
Hmmm, seems like it is an option in the workflow format:
https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull-request-events-for-forked-repositories

Seems like `pull_request` is the missing trigger.
I think perhaps a change like this in `.github/workflows/commit.yml` should be good enough, but we won't really know until it has been merged to master.
```
on: [push, pull_request]
```

It would probably be best to fix this issue before merging this PR

trym-b:
Created a PR for this: https://github.com/caveman2cosmos/Caveman2Cosmos/pull/809

MattOttawa:
Looks good man.

MattOttawa:
Ya, very nice!

trym-b:
Happy to help.

In fact, it was quite fun to get this to work, and it makes me happy that you have already started trying this system out. Let me know if there are any other tasks for such a fresh contributor as me to help with, perhaps I can find the time to help out some more.

Also: you might want to consider to require both of the pylint tests to pass before merging to master, this way people won't be able to accidentally merge detectable python regressions.

MattOttawa:
I've had a dream to make some ipc for c2c ever since I heard of it. I've done a bit of reading and I found some example code to copy paste, but I'm in over my head with this stuff for sure. I'll make a pr so you and others can see and maybe help.

dzfischer:
lol... yeah, I expect this to flag a number of things in the short scripts I wrote, they're mostly just hackish ways to get a better visualization of gamedata or the like. Everything in Blazescripts is gonna be like that; anything the game itself uses should be python2 though, and this def could be useful there.

PR #823, opened 2021-01-18
enabled python error: bad-whitespace

MattOttawa:


MattOttawa:
I don't think we want this one enabled.

PR #827, opened 2021-01-18
enabled python error: used-before-assignment

MattOttawa:


MattOttawa:
************* Module MapScriptTools
E: 85, 9: Using variable 'bInitialized' before assignment (used-before-assignment)
************* Module BarbarianCiv
E:148, 8: Using variable 'CyTeam' before assignment (used-before-assignment)

PR #836, opened 2021-01-18
enabled python error: assert-on-tuple

MattOttawa:


MattOttawa:
************* Module CvUtil
W:108, 1: Assert called on a 2-uple. Did you mean 'assert x,y'? (assert-on-tuple)

PR #846, opened 2021-01-18
enabled python error: pointless-statement

MattOttawa:


MattOttawa:
************* Module PediaImprovement
W:188, 3: Statement seems to have no effect (pointless-statement)
************* Module CvMainInterface
W:1194, 4: Statement seems to have no effect (pointless-statement)
************* Module C2C_World
W:3299, 3: Statement seems to have no effect (pointless-statement)

Toffer90:
Ok, only one of them was wrongly corrected in this PR, the one in the pedia needed to be deleted as it would break stuff if added to the PF string.

PR #847, opened 2021-01-18
enabled python error: bad-super-call

MattOttawa:


MattOttawa:
************* Module BugOptions
E:1192, 2: Bad first argument 'LinkedOption' given to super() (bad-super-call)

PR #854, opened 2021-01-29
CI: Downgrade to Ubuntu 18.04 to avoid get-pip

trym-b:
There have been a recent update to the `get-pip.py` script for
installing `pip` for python 2.7. This updated causes installing pip to
fail with the following error message:
```
Traceback (most recent call last):
  File "get-pip.py", line 24226, in <module>
    main()
  File "get-pip.py", line 199, in main
    bootstrap(tmpdir=tmpdir)
  File "get-pip.py", line 82, in bootstrap
    from pip._internal.cli.main import main as pip_entry_point
  File "/tmp/tmpG8aJb5/pip.zip/pip/_internal/cli/main.py", line 60
    sys.stderr.write(f"ERROR: {exc}")
                                   ^
SyntaxError: invalid syntax
```

This commit downgrades the Ubuntu version from 20.04 to 18.04 to avoid
using this script. Ubuntu 18.04 can install pip from the default repo,
avoiding `get-pip.py` entirely.

Another benefit is that it simplifies the setup script as well.

### DIFF
--- a/.github/workflows/commit.yml
+++ b/.github/workflows/commit.yml
@@ -18,7 +18,7 @@ jobs:
                       --rm \
                       --interactive \
                       --volume ${{ github.workspace }}:/host_dir \
-                      --workdir /host_dir ubuntu:20.04 \
+                      --workdir /host_dir ubuntu:18.04 \
                       bash -c \
                       "./Tools/Lint/setup.sh && python3 Tools/Lint/pylint.py --python-versions ${{ matrix.python-version }}"
     python-unit-tests:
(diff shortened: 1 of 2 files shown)


MattOttawa:
Nice, thanks. I wasn't sure what happened or how to fix.

MattOttawa:
I also don't know how to disable merging prs if check don't pass like you suggested earlier.

trym-b:
I think it is fine to not have it required, since it was not stable until now. Also, no one has actually pushed anything that has violated any of the currently enabled `pylint` errors. If that happens I think we could consider it again.

PR #856, opened 2021-01-30
CI: Run XMLValidator in Github Actions

trym-b:
Since knowing that all XMLs are valid is critical for a successful
build, all of the XMLs in the repo are now checked on each push/PR to
the repo from the outside.

Hopefully this will make it easier for external contributors to know
that their XMLs are valid before a potential merge. The reason for this
is that although the validator is currently run in `Appveyor` today, it
is not easy to replicate the setup for getting `Appveyor` to run on a
forked repo. With `Github Actions`, all of this is handled for the user.

One drawback with this approach is that with this commit merged the
validation is run twice. But the benefit of reducing the chance of
breaking master should outweigh this drawback.

### DIFF
--- a/.github/workflows/commit.yml
+++ b/.github/workflows/commit.yml
@@ -38,3 +38,15 @@ jobs:
                       --workdir /host_dir ubuntu:18.04 \
                       bash -c \
                       "./Test/Python3/setup.sh && pytest -c Test/Python3/pytest.ini ."
+    validate-xmls:
+        name: Validate XMLs
+        runs-on: windows-2019
+        steps:
+            - name: Check out code
+              uses: actions/checkout@v2
+              with:
+                  lfs: true
+            - name: Run XmlValidator
+              working-directory: ${{ github.workspace }}/Tools
+              run: |
+                  ${{ github.workspace }}/Tools/XmlValidator.exe -a


Toffer90:
Nice one.
XML validator is fast, (about a minute), so running it twice doesn't cost much.

trym-b:
Would you also like to build to code through Github Actions as well? Although I am not suggesting to drop `Appveyor` (because of the SVN stuff, etc), however I am currently under the impression that it is much easier to work with GA compared to Appveyor.

My main issue with Appveyor is that I simply do not know how to reproduce the Appveyor build on external forks, which would force external contributors to build the code manually to verify that everything builds before making a PR. With GA it is just as simple as to fork the project and there you go, you have already a working CI setup.

What do you think? It the resulting workflow would look something like this: https://github.com/trym-b/Caveman2Cosmos/runs/1795310010

Toffer90:
Don't have much of an opinion on this as I don't work with a fork.
This guy (@billw2019:) set up appveyor for us, so he's better suited to discuss this further.

billw2012:
IMO (for what it is worth since I left this project more than a year ago), you should definitely move to github actions. They were only just coming on line when I was last working on this, and I did plan to move everything over to them. I have used them for testing and deployment on other projects and they are far superior to Appveyor.

PR #864, opened 2021-02-03
deleted unused parameter and added const

MattOttawa:


Toffer90:
Looks good, but should wait till after v41.1 is released.

PR #865, opened 2021-02-04
turned the Great Wall graphics on

MattOttawa:


flabbert:
The great wall could create a fatal error is that fixed?

Toffer90:
Not sure, there was a bad issue if the great wall ended up circumnavigating the world.

It also had some visual oddities when playing with viewports as it had to be refreshed every time viewport changed, and it then refreshed along the current borders rather than staying at the borders as they were when it was built as is how it visually works without viewports.

MattOttawa:
Isn't it the option that makes it do that?
Sounds like deleting the option and what it does will fix the issues.

	<Define>
			<DefineName>DYNAMIC_GREAT_WALL</DefineName>
		<bDefineBoolVal>1</bDefineBoolVal>
	</Define>

PR #866, opened 2021-02-05
Interesting to have the change comparison between the branches.

Toffer90:
Buildings that obsolete are removed from city completely, and replaced buildings use the disabled building system rather than having its own system.
This is still far from finished; a lot of quality control work left.
I also need to add new obsoletion building version of most wonders that obsolete at a minimum (I added a tag that indicate what building should be built when this building obsoletes, I imagine mostly wonders using this, but there may be other buildings it makes sense for).
Have to double check any code that asks if a building is obsolete, as that may now be redundant if it also asks if the city has the building.


raxo2222:
There may be problem with requirements obsoleting before building itself obsoletes.

Is it possible to ensure, that you can build thing despite all its building requirements being obsolete?

For example X requires A, B, C OR D.
Those four never got built, and got obsoleted.
Then you still could build X.

Otherwise all building requirements of buildings need to be revised.

Toffer90:
> There may be problem with requirements obsoleting before building itself obsoletes.
> 
> Is it possible to ensure, that you can build thing despite all its building requirements being obsolete?
> 
> For example X requires A, B, C OR D.
> Those four never got built, and got obsoleted.
> Then you still could build X.
> 
> Otherwise all building requirements of buildings need to be revised.

I've thought about it, and decided that I'll make it so that if a required building is obsolete then the requirement is not needed to be fulfilled.

raxo2222:
> 
> I've thought about it, and decided that I'll make it so that if a required building is obsolete then the requirement is null and void and will be ignored. One could say it is assumed that if a building is obsolete, all cities have the building without technically having it.

Nice, this will help with new cities especially in later eras, as then mistimed requirement obsoletions are worst.

raxo2222:
Wont those <ObsoleteSafeCommerceChanges> duplicate with <CommerceChanges>?
In some cases those could be far from each other in building define

Toffer90:
> Wont those <ObsoleteSafeCommerceChanges> duplicate with <CommerceChanges>?
> In some cases those could be far from each other in building define
> 
> <CommerceChanges> - 3767 in master, and 4157 in this branch.

Not sure what you mean, I changed ObsoleteSafeCommerceChanges to CommerceChanges; if a building already had CommerceChanges I took the values from the obsolete one and added it to the regular one.

Toffer90:
@raxo2222 feel free to playtest this branch and give me feedback, I haven't had much time to properly test things in-game yet other than to see that the obvious things apparently work.

raxo2222:
> 
> 
> > Wont those  duplicate with ?
> > In some cases those could be far from each other in building define
> >  - 3767 in master, and 4157 in this branch.
> 
> Not sure what you mean, I changed ObsoleteSafeCommerceChanges to CommerceChanges; if a building already had CommerceChanges I took the values from the obsolete one and added it to the regular one.
> 
> Almost all wonders and religious buildings had ObsoleteSafeCommerceChanges and almost none of them had CommerceChanges, so it's not strange that there are a lot more CommerceChanges in this branch than master at this point.
Oh, that's why more of them appeared.

PR #869, opened 2021-02-07
Research long long

MattOttawa:


flabbert:
what is the status of this? it feels like an unfinished issue?

MattOttawa:
This breaks saves. It's possible to switch so it doesn't, but I always seem to screw up the save/load, so I'd rather to wait till save break.
If any1 wants to make this save game compatible, go ahead.

flabbert:
lets close this PR for now, and revisit this issue if its required. It is very likely that we will have changed other bits of the code.

We keep the branch, and create an Issue where we list the branch.

PR #872, opened 2021-02-07
fixed assert

MattOttawa:
CvPlayerAI::AI_getAttitude
ePlayer != getID()
shouldn't call this function on ourselves


Toffer90:
I've already fixed this one in my branch.

MattOttawa:
The upcoming release probly won't be better than the last release.
When is the freeze over? I'm gonna race to merge my prs first so that others get the conflicts.

Thunderbrd:
Next release in a week or two as I'm dusting off the old sea legs to do some debugging on some known issues before releasing a bug fix version here.

PR #875, opened 2021-02-09
deleted CvXMLLoadUtility::RemoveTGAFiller

MattOttawa:
GC.getGAMEFONT_TGA_RELIGIONS() is 50 right now. This does nothing I think. MRGENIE stuff is usually complicated but pointless.


Toffer90:
Its value is very important in how the tga symbol texture is read.

MattOttawa:
Have you looked? I just searched. It's more pointless bs.

Toffer90:
Changing the value of that global define does have a noticeable effect in-game, I changed it while I reworked the GameFont.tga files. I think its old value was 100, which was way too much and caused the game to have too few unicodes for symbols left for all the resource symbols as too many empty ones were reserved for religions and corporations.

MattOttawa:
Hmm. I was searching the tags.

Toffer90:
Texture modders usually never touch cpp code, so we should keep the xml define rather than hardcoding it.

PR #878, opened 2021-02-14
added map category check for barb city spawn

MattOttawa:


raxo2222:
Can be this added to V41.1 release?

Toffer90:
> Can be this added to V41.1 release?

Sure, it's the same as I did for regular barb city spawn.
This is only for the cities created before turn 0 by barbarian world, I didn't realize that those cities are founded by a different code than the normal barb city spawn back then.

PR #887, opened 2021-03-07
feat(Language): German translations

dsmabell:
Changed Group titles to the use of variables to support german grammar (CivGame.cpp, GroupTitle...xml)
Correction in CivOutcome.cpp to get correct grammar of outcome
more animal translations
translations to some combat modifiers
correction and unification of SHIFT translation


dsmabell:
Review of CivGame.cpp code changes required

raxo2222 (commented):
[UMSCHALT] is not a thing - essentially you now have word in square brackets instead of reference to actual thing that is [SHIFT]

That is [] is used for text formatting, otherwise game may be confused.

PR #889, opened 2021-03-09
Mod created three different file inside of game folder instead of mod…

raxo2222:
… folder

I'm not sure if I changed things correctly, but files are now created in mod.
Memory log directly in Caveman2Cosmos, and two map setting files inside of user settings.


raxo2222:
memory.log has currently hardcoded path in cpp files, meaning it would create Caveman2Cosmos folder in Mods regardless of actual mod folder.

Maybe outright force users to rename mod folder to Caveman2Cosmos?

PR #902, opened 2021-03-21
Added Afflictions feature macro

MattOttawa:


flabbert:
I would love a description of what this actually is? Is it a new game concept?

MattOttawa:
OUTBREAKS_AND_AFFLICTIONS is one of TB's combat mods. It is incomplete.
I added #ifdef OUTBREAKS_AND_AFFLICTIONS #endif around all the code so it can be turn on/off easily.
It breaks saves to turn it off, so I left it on - which means this PR does 100% nothing.

PR #920, opened 2021-04-14
switched getInfoTypeForString

MattOttawa:


Toffer90:
Didn't notice this copy/paste irregularity, did the colour coding in a different mod, so just copied it over.
Nothing gets past you, nice.

PR #922, opened 2021-04-15
Xml cleanup

raxo2222:
I cleaned up whitespaces and comments there


Toffer90:
Things like this doesn't need to be a PR, could just push it straight to master, nothing controversial or potentially risky in white space changes.

raxo2222:
Okay :)

Toffer90:
Just don't do this kind of mass edit thing on the source files, they are usually structured with some redundant white spaces on purpose for readability reasons.

raxo2222:
I know, did this only on XML files

PR #928, opened 2021-04-18
Worker/evaluation fix

flabbert:
This PR looks to improve worker evaluation, give AI a better chance at keeping up in the earlygame, and remove the farm is being replaced by seed gatherer bug

### DIFF
--- a/Sources/CvCityAI.cpp
+++ b/Sources/CvCityAI.cpp
@@ -232,6 +232,7 @@ void CvCityAI::AI_doTurn()
 		AI_stealPlots();
 	}
 
+	AI_markBestBuildValuesStale();
 	AI_updateBestBuild();
 	AI_updateWorkersNeededHere();
 
@@ -8215,13 +8216,13 @@ int CvCityAI::AI_getImprovementValue(CvPlot* pPlot, ImprovementTypes eImprovemen
 
 	if (eBonus == NO_BONUS)
 	{
-		for (int iJ = 0; iJ < GC.getNumBonusInfos(); iJ++)
-		{
-			if (improvement.getImprovementBonusDiscoverRand(iJ) > 0)
-			{
-				iValue += 10;
-			}
-		}
+		// for (int iJ = 0; iJ < GC.getNumBonusInfos(); iJ++)
+		// {
+		// 	if (improvement.getImprovementBonusDiscoverRand(iJ) > 0)
+		// 	{
+		// 		iValue += 0;
+		// 	}
+		// }
 	}
 	else if (!GET_TEAM(getTeam()).isBonusObsolete(eBonus))
 	{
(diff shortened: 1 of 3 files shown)


Sources/CvCityAI.cpp
```diff
@@ -232,6 +232,7 @@ void CvCityAI::AI_doTurn()
 		AI_stealPlots();
 	}
 
+	AI_markBestBuildValuesStale();
```
Toffer90, line 235: Pretty sure it is not needed to mark it as stale every turn for all cities.
It is marked as stale for each building that is built or removed, for each culture level change, and every time an improvement change inside the city's workable radius, every time a civic is changed, and for each tech invented.
That should be more than often enough.

flabbert: You are very probably right, It was part of an early attempt to see if i could find an improvement, it didnt fly that well, but i havent removed it yet.

Sources/CvCityAI.cpp
```diff
@@ -8475,13 +8477,13 @@ void CvCityAI::AI_updateBestBuild()
 					}
 				}
 
-				if (aiFinalYields[YIELD_FOOD] >= iFoodPerPop)
+				if (aiFinalYields[YIELD_FOOD] >= iFoodPerPop) // if total food yield (on all plots?) is higher than food needed for 1 plot
```
Toffer90, line 8480: No, this is only the yield for the current plot in the plot loop.
So `if (plot has equal or more food yield than 1 population consume) { do something; }`

Sources/CvUnitAI.cpp
```diff
@@ -1956,7 +1955,7 @@ int CvUnitAI::AI_minSettlerDefense() const
 		*
 		(
 			// One defender if game-start settler(s), else 4.
-			1 + 3 * (getGameTurnCreated() > 0)
+			1 + 1 * (getGameTurnCreated() > 0)
```
Toffer90, line 1958: Should update comment to reflect code, `else two`.

Sources/CvUnitAI.cpp
```diff
@@ -2103,7 +2101,7 @@ void CvUnitAI::AI_workerMove()
 	{
 		const int iWorkers = GET_PLAYER(getOwner()).AI_totalUnitAIs(UNITAI_WORKER);
 
-		if (iWorkers > 3 && iWorkers > GET_PLAYER(getOwner()).getNumCities()
+		if (iWorkers > 3 && iWorkers > 5 * GET_PLAYER(getOwner()).getNumCities()
```
Toffer90, line 2104: Since this is in the section where AI is in serious financial trouble, then it should not need 5 workers per city.

Toffer90:
Other than what I commented on, I don't see anything that worries me in this PR.
If you in your test game see or feel that it has the desired effect, then thumbs up.

Sources/CvUnitAI.cpp
flabbert: what defines as "serious financial trouble?"

Toffer90: That expenditure is 85% or more than than the entire gross income when at 100% tax rate (0% science, culture espionage)
But not always when that is the case either, there's more rules to it than just that, like if tax rate is more than 50% and the current treasury cannot fund you for X amount of turns at the current rate (X differ by gamespeed, 10 turns on normal gamespeed); i.e. not hit 0 gold treasury within X amount of turn.

It's more complex than that though, but it is pretty bad when that function call returns true.
Leader personality also influence at what threshold of bad it consider to be financial trouble.

PR #948, opened 2021-04-19
deleted static from CvPlot::rect

MattOttawa:
Looks better

### DIFF
--- a/Sources/CvPlayer.cpp
+++ b/Sources/CvPlayer.cpp
@@ -2619,10 +2619,9 @@ void CvPlayer::acquireCity(CvCity* pOldCity, bool bConquest, bool bTrade, bool b
 
 	if (bTrade)
 	{
-		foreach_(CvPlot* pLoopPlot, CvPlot::rect(pCityPlot->getX(), pCityPlot->getY(), 1, 1))
-		{
-			pLoopPlot->setCulture(eOldOwner, 0, false, false);
-		}
+		algo::for_each(pCityPlot->rect(1, 1),
+			bind(CvPlot::setCulture, _1, eOldOwner, 0, false, false)
+		);
 	}
 
 	CvCity* pNewCity = initCity(pCityPlot->getX(), pCityPlot->getY(), !bConquest, false);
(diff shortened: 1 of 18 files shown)


Sources/CvPlayer.cpp
```diff
@@ -2619,10 +2619,9 @@ void CvPlayer::acquireCity(CvCity* pOldCity, bool bConquest, bool bTrade, bool b
 
 	if (bTrade)
 	{
-		foreach_(CvPlot* pLoopPlot, CvPlot::rect(pCityPlot->getX(), pCityPlot->getY(), 1, 1))
-		{
-			pLoopPlot->setCulture(eOldOwner, 0, false, false);
-		}
+		algo::for_each(pCityPlot->rect(1, 1),
+			bind(CvPlot::setCulture, _1, eOldOwner, 0, false, false)
+		);
```
flabbert, line 2624: this means that only the plot that city is on gets culture swapped? 
setCulture should be investigated further so that culture around the city get atleast partially flipped to conquering civ

Sources/CvPlayer.cpp
```diff
@@ -6401,7 +6400,7 @@ void CvPlayer::receiveGoody(CvPlot* pPlot, GoodyTypes eGoody, CvUnit* pUnit)
 			int iBestValue = 0;
 			pBestPlot = NULL;
 
-			foreach_(CvPlot* pLoopPlot, CvPlot::rect(pPlot->getX(), pPlot->getY(), iOffset, iOffset)
+			foreach_(CvPlot* pLoopPlot, pPlot->rect(iOffset, iOffset)
 			| filtered(!CvPlot::fn::isRevealed(getTeam(), false)))
 			{
```
flabbert, line 6405: It would be helpful when using these binds that you add some comment with what is happening, so other modders can follow the code down the line

flabbert:
I added a couple of comments, also, it would be hugely beneficial if the rect function had a descriptive comment over it so that it can be read when hovered in some editors

Sources/CvPlayer.cpp
MattOttawa: There is pinned message in cpp channel. I couldn't explain better then bill or google.

Sources/CvPlayer.cpp
MattOttawa: There is pinned message in cpp channel. I couldn't explain better then bill or google.

flabbert: Sure i know that now, but new modders that come to later maybe wont

PR #961, opened 2021-04-28
disabled unused MAD nukes code

MattOttawa:
I'd be happy to just delete this code if nobody cares.


raxo2222:
Feel free to delete it - nobody will miss feature, that didn't exist for long time anyway.

Thunderbrd:
I spoke with the creator of this system a while back and the people that eliminated his work on this didn't realize they were doing it at the time they did.  It sounds like it would be a neat feature but I never did try to get it running.  I'd suggest someone try it out before just getting rid of it, though I don't know how specific it might be with certain things.  It was supposed to give some interesting detail during the nuclear high alert eras in the military advisor.

raxo2222:
So tag needs to be added to some civic, so it can be activated with civics.

Thunderbrd:
Yeah apparently but now its commented out and I'm not sure how exactly it works when it does - we removed the MAD civic as well and it would really only be appropriate when you're in a Mutually Assured Destruction strategic mode of some sort, which he keyed into that civic.

flabbert:
write down the ideas on how it was supposed to work, and then nuke it (pun intended), then the concept can be revisited if and when we have the tools in place to do it properly

Thunderbrd:
It functions fine - I have no idea what it was supposed to do just remember it sounded cool af.  It would be good if we just turn it on and test it and see what it does and then figure out how to best incorporate it at some point.  Until someone can spend some time looking into it, keeping it off seems appropriate.  I think most of the coding is purely python anyhow.

flabbert:
I have made a ticket in the "cool features to add again later" list, no reason to have this PR lying here ad infinitum, lets remove it for now, the amount of AI rework that we have in progress atm anyway means that it probably will be easier to readd it later, and make it work properly (if we still want it, clearly we didnt miss it that much)

PR #967, opened 2021-04-29
Added Game Option tab to BUG options screen

MattOttawa:


Toffer90:
This is cool, though this tab should perhaps be hidden if not in debug mode, as many of these options are not safe to change mid-game, as all other BUG options are.
At some point we may add a  gameoption boolean tag "bSafeToChangeMidGame", so that this BUG tab doesn't have to hide them all.
Anyhow, just a couple suggestions, not important really.

MattOttawa:
Done adding tag. I'll leave the xml up to Raxxo.

Toffer90:
Size matter is the only one I know for sure doesn't like being changed mid-game, but I'm sure there's others.

PR #1005, opened 2021-06-01
Timesort - maybe make it some sort of debug mode?

raxo2222:
It would be nice to have switch to see when stuff is unlocked directly in pedia:
Techs display tech location. 
Units also display such information - was already present, but I had to fix bug with no tech units.
Some infotypes like units, buildings, civics, promotions, builds, improvements, bonuses have xgrid appended to their names in pedia list.
Buildings are sorted by highest tech requirement.
Units are assumed to be sorted by highest tech requirement, bonuses are sorted by tech reveal.
Other sorted stuff have single tech requirement.


raxo2222:
No need to keep this open anymore, though ability to sort pedia entries by techs would be nice

PR #1007, opened 2021-06-04
switched vector to IDValueMap

MattOttawa:
also did:
// Toffer - ToDo - change to iterate the techs cached in the building vector rather than iterate through all techs to see if they are in the vector.
// TODO reform loop to iterate on the mappings


Toffer90:
Nice.

PR #1016, opened 2021-06-06
switched arrays that use global defines to vectors

MattOttawa:


Toffer90:
Haha, that was fast, nice. I had a feeling you probably had tinkered with this on your end already.

PR #1024, opened 2021-06-10
added delete for m_commander and added CvUnitComponents to vs files

MattOttawa:


Toffer90:
So, am I right in assuming that c++ garbage collector doesn't clean up allocated object instances when there no longer exist a pointer pointing to it?
That explicit delete is needed to avoid memory leak in this case?

MattOttawa:
Ya, only when new is used to create the obj. I think we can get rid of the new and make it m_commander = &UnitCompCommander();  but im only 99% sure, so i did this. Maybe I change.

MattOttawa:
Nevermind. this makes more sense.

PR #1026, opened 2021-06-12
switched a loop for unit build types to foreach_

MattOttawa:
Matt: this loop
// Toffer - ToDo
//	- Should limit loop to only UnitInfo builds and the extra builds for this unit.
//	- Both are vectors and are both referenced in the CvUnit::hasBuild() function.

Matt: switched this to a boost function, but it's probly not much faster.
// Toffer - This can't be faster than just doing an erase of the element with the build id; can it?

### DIFF
--- a/Sources/CvInfos.h
+++ b/Sources/CvInfos.h
@@ -2244,9 +2244,10 @@ public:
 
 	bool isPrereqOrCivics(int i) const;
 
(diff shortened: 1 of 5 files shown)


Toffer90:
Nice, love that you pick up on my comments about stuff I save for later as a todo.

PR #1044, opened 2021-06-26
switch default for max units/buildings from -1 to 0 to fix FAssert

MattOttawa:


Toffer90:
Good call, no reason to have the value `0` mean building/unit is Disabled`, there's other more proper ways to disable them.

PR #1086, opened 2021-07-22
changed path code for TechSearcher.py

MattOttawa:


raxo2222:
It finds file but somehow its unable to process it

raxo2222:
So it seems like it was written in Python 2 - it never worked in Python 3 as I tried using original file

PR #1090, opened 2021-07-24
Parallel maps unit mission

MattOttawa:


MattOttawa:
maybe I delete some of this and use outcome missions

raxo2222:
If those can be given tech requirements, then it should work
None of this should be available before Satellites tech for units

Thunderbrd:
Why shouldn't missions be available into prehistory for entering subterranean maps through some select 'empowered' cave features?

raxo2222:
I was talking about "send to space" missions.

As for subterranean map there could be one for Earth, Moon, Mars, Venus, and Titan.
This would mean defining 5 more maptypes in xml and code, also needing at least 5 more terrains

Thunderbrd:
Yeah I've thought about that - might be a matter of how much we can data manage multimaps but if it works out to have those I'm not against them.

PR #1139, opened 2021-08-29
Myth story stories rework

raxo2222:
Removed Myth Effects, Storys, Stories and Stories effects.

This means city screen is decluttered, also no 400 entries (200 storys and stories) in bard and story teller
Myths give themselves now.

Now it needs some new tag set


MattOttawa:
tags to do what?

raxo2222:
Percentage commerce changes for tech would be very nice - its a thing for bonus commerce tags, so similar concept could be applied there

For example myth would give +1 research, and then it could give +0.9 research and 0.1 culture with later tech and so on.

PR #1140, opened 2021-08-30
modified IDValueMap so value can be an array

MattOttawa:


MattOttawa:
I need xml and schema for new tag to see if this works. You do that raxxo?

raxo2222:
Yeah

raxo2222:
Those are fractional tech commerce changes - that is +x or -x to value
Not +-% modifiers

raxo2222:
I think it would be easier if you first converted bonus commerce percents to idvalue map, and then wrote tech commerce percents in idvalue form

PR #1144, opened 2021-09-06
added max unit check to outcome missions

MattOttawa:


flabbert:
is this still relevant?

PR #1152, opened 2021-09-12
code that can help adding/changing xml tags in the dll

MattOttawa:
Tags that are added to CvBuildingInfo::getDataMembers() (like the example) won't need any code added to the checkSum, copyNonDefaults and readXml functions.

Todo - Need more code to handle different tags/data types.


Toffer90:
Cool initiative.

PR #1155, opened 2021-09-13
wrapped 2 array with boost::scoped_array

MattOttawa:


Toffer90:
What does this achieve?

MattOttawa:
Ya. Don't need that there though. I change.

PR #1156, opened 2021-09-14
added value to var incase it is not saved

MattOttawa:
just a guess. maybe it's not needed.


Toffer90:
> just a guess. maybe it's not needed.

I'm not sure it's needed or not, but it's a good guess.

raxo2222:
> 
> 
> > just a guess. maybe it's not needed.
> 
> I'm not sure it's needed or not, but it's a good guess.

It was needed after all - there was uninitialized variable issue, that broke building terrain yields

PR #1168, opened 2021-09-18
deleted free promotion tags

MattOttawa:


MattOttawa:
I got double text in pedia now? Someone check and tell me what to delete please.

Issue #469, opened 2020-03-06
Convert nested boolean tags to vectors


raxo2222:
There are some nested tags, that contains booleans to turn them off/on.
They can be removed/vectorized in code/xml.
Count Type
5065 \<bUnitAI>
2720 \<bGreatPeople>
2549 \<bUnitCombat>
1904 \<bBuild>
1890 \<bFreePromotion>
1110 \<bTerrain>
926 \<bMakesValid>
812 \<bTrait>
530 \<bLeaderType>
426 \<bUnitCombatTarget>
425 \<bPrereqTerrain>
338 \<bAttitudeType>
331 \<bPrereqCivic>
285 \<bPrereqFeature>
216 \<bFeatureTerrain>
203 \<bFeature>
185 \<bDiplomacyPowerType>
152 \<bLeaderAvailability>
44 \<bValid>
41 \<bTerrainDoubleMove>
27 \<bFeatureDoubleMove>
25 \<bUnitCombatDefender>
21 \<bHurry>
20 \<bTerrainTrade>
5 \<bFreeTech>
4 \<bNotRequired>
1 \<bDisableTech>
1 \<bCivicOptionNoUpkeep>

raxo2222:
@alberts2 Do you want to do it? 
These shouldn't break saves.

alberts2:
Only if it really makes sense to do it. I don't remember anything that suggests that change would yield a real improvenent.

raxo2222:
Weird, I remember, when someone said, that those booleans should be vectorized for performance gain or memory savings...

alberts2:
The memory saving would be tiny and in some cases the performance could be worse than it is now. One reason for that is that rsearching something in a vector is slower than simply returning a bool from a array.

MattOttawa:
Looking at <bUnitAI>it's like what Alberts said. Not sure if getting rid of bool is better.
It is easy to add a default when this stuff is read so you could get rid of the bool tag, just not the extra nesting.

Toffer90:
There's so few different UnitAI's that they can just as well be arrays containing one element per possible UnitAI.
More important to vectorize the data that contains more than a 1000 elements when only less than a 100 of them actually have a valid value.

Thunderbrd:
That's a good point - this is mostly about the stuff that has grown quite large.  The boolean part isn't the problem it's the large array part.

raxo2222:
I updated this reference

flabbert:
[...]

PR #1177, opened 2021-09-28
switched CvBuildingInfo::PrereqOrImprovement from array to vector

MattOttawa:


MattOttawa:
#469
Wanna do the xml and schema for this raxo?

PR #1192, opened 2021-10-09
switched vector to IDValueMap

MattOttawa:


flabbert:
is this still relevant?

PR #1205, opened 2021-10-23
Rpc

MattOttawa:


MattOttawa:
I've added a new exe to run a second process.
There is multiple ways to create the required IPC.
[https://www.geeksforgeeks.org/inter-process-communication-ipc/](url)

I use RPC here. (recommended by Overmind)
[https://en.wikipedia.org/wiki/Remote_procedure_call](url)

.idl files are compiled by a MIDL compiler to build an RPC interface.
VS does the compiling, so we just need to write .idl files for calls that go from the dll to the new exe.-
[https://en.wikipedia.org/wiki/Interface_description_language](url)

Each .idl file will generate one .h and two .c files when compiled.
These files will be included in the dll and exe when they are compiled.

MattOttawa2:
My dream is to move all the current code to the new process.
Then the old dll would have classes and functions that match the DllExports (and maybe the python calls) to bounce calls between the new process and the old exe.

Toffer90:
> My dream is to move all the current code to the new process. Then the old dll would have classes and functions that match the DllExports (and maybe the python calls) to bounce calls between the new process and the old exe.

It's a great dream, a dream we share. ^^

PR #1216, opened 2021-11-09
Yield ratio and value refactor

flabbert:
this PR reworks how yield values are calculated and stored. 

This is the initial work to put yield ratios and value into a struct, and have the ability to modify this struct until calculating the value of a plot based on it in the end.

It only uses base predetermined values for now, but it is built in such a way that it can be called to modify weights


Toffer90:
lol
image

flabbert:
yeah that one was amusing, i wont doubt there are more  reference issues around either.

PR #1232, opened 2021-12-21
Rework best plot build again

flabbert:
this reworks AI_newbestplotbuild (to be renamed to AI_bestplotbuild) again

the function specifically checks what is the best improvement and build to use for plots for a city, and only checks plots inside the city limit of that city.


Sources/CvCityAI.cpp
```diff
@@ -10922,30 +10922,26 @@ void CvCityAI::AI_newbestPlotBuild(const CvPlot* pPlot, plotInfo* plotInfo, int
...
-		{
-			bValid = true;
-		}
-		//check if improvement is a fort or watchtower, then its a no.
-		else if (!potentialImprovementInfo.isActsAsCity() && potentialImprovementInfo.getVisibilityChange() == 0)
-		{
-			int iValue = 0;
-			int iBestTempBuildValue = 0;
-			if (potentialImprovementInfo.getNumBuildTypes() > 1) continue;
+		// check if improvement is a fort or watchtower, then its a no.
```
jonask84, line 10932: Maybe separate each of these checks with a blank line for readability

PR #1234, opened 2021-12-22
Used CvImprovementInfo build types cache in a couple more spots

MattOttawa:


raxo2222:
Can <CreatingBuilds> removed after this? Or is it separate thing
This tag isn't needed once we have proper caching.

MattOttawa:
I'll do that next.

PR #1235, opened 2021-12-22
added code to create CvImprovementInfo build types cache

MattOttawa:


Sources/CvCityAI.cpp
```diff
@@ -10936,19 +10936,18 @@ void CvCityAI::AI_newbestPlotBuild(const CvPlot* pPlot, plotInfo* plotInfo, int
 		// if current improvement is same as potential improvement we dont need to reevaluate
 		if (ePotentialImprovement == eCurrentPlotImprovement) continue;
 		// if more than 1 build can build this improvement, find fastest
-		if (potentialImprovementInfo.getNumBuildTypes() >= 1) {
```
MattOttawa, line 10939: @flabbert billw told me I didn't need the line above in a similar situation. I'm guessing you don't mind if I delete it.

flabbert: no, its fine to remove it, it was a copy paste from me originally that didnt work at all, so i added >=

You are absolutely correct, if the list is empty it wont iterate through it and wont generate a value, so the if would be redundant and should be removeed

flabbert: I welcome any improvements to the code I have done here, it is not perfect yet, and there are edge cases not properly thought about. (Particularly when to remove features is something that can be vastly improved upon, try to keep certain amount of forest around a city for later improving for instance

MattOttawa:
the CreatingBuilds tag can be deleted now.

PR #1241, opened 2021-12-26
added try/catch around python::stop_iteration_error()

MattOttawa:


MattOttawa:
I assume you are still getting those exceptions Toffer? 
My first fix attempt failed.
This next attempt might not be a proper fix, but it will give me time if it works, and I need to confirm my guessing anyways.

Toffer90:
> I assume you are still getting those exceptions Toffer? My first fix attempt failed. This next attempt might not be a proper fix, but it will give me time if it works, and I need to confirm my guessing anyways.

Yeah, haven't checked in a couple days, but I got it right before christmas.
Thanks for investigating it, is real annoying to get those exceptions when debugging other stuff.

Issue #1275, opened 2022-01-09
Windows 7 crashes


MattOttawa:
Sometimes changes work on my computer and crashes on raxos.
Some investigation might help to avoid this problem in the future.

https://github.com/caveman2cosmos/Caveman2Cosmos/pull/1270/files

MattOttawa:
Apparently adding a bit of unused code can crash windows 7?https://github.com/caveman2cosmos/Caveman2Cosmos/commit/ed30cdcc77b582229c4446266bdc527bf9a082ce

raxo2222:
Yes, no idea why it was crashing with that, now its solid as rock

MattOttawa:
post a mini dump next time if you get one please.
Or better yet - attach the VS debugger and post screen shots.

Toffer90:
Yeah, since you're the only dev with windows 7, it might be practical if you learn how to attach VS to C2C.

Just install base VS community, barebone install (no components selected), and open "...\Sources\C2C" (VS2019).sln" with it.
It will tell you to install at least one component. After that you just need to select x86 in the dropdown box where it says "any CPU", the dropbox is displayed close to VS toolbar at the top.
After that you simply hit the green play button, and VS will launch the game.

MattOttawa:
art info enums part 1 #1303

raxo2222:
Didn't get crashes from it so its all good

MattOttawa:
#1316

raxo2222:
No crashes here

MattOttawa:
#1317

raxo2222:
No problems here

PR #1277, opened 2022-01-10
CvOutcomeMission tweaks

MattOttawa:
Can you test this raxo? this is about half of what you reverted.

#1275


raxo2222:
Yeah, this one works

PR #1279, opened 2022-01-10
modified the shrine buildings cache

MattOttawa:
the cache is created from xml values and doesn't change I think, so I set it up like the other xml based caches.


Toffer90:
Good catch, it's utterly pointless to cache that xml data every turn and save it in saves... Who wrote that car wreck? lol.

MattOttawa:
lol looks like post bts and pre c2c code to me.

Toffer90:
> lol looks like post bts and pre c2c code to me.

Hehe, yeah, the early days, when inexperienced modders were superstitious and saw ghosts around every code corner. 
This code seem so pointless that there has to have been some kind of spiritual justification for its design. lol.

Issue #1164, opened 2021-09-18
Convert <GlobalBuildingExtraCommerces> to normal 2D tag in XML


raxo2222:
Currently it is only one with weird XML formatting:

\<GlobalBuildingExtraCommerces\>
  \<BuildingExtraCommerce\>
     \<BuildingType\>BUILDING_TAILOR_SHOP\</BuildingType\>
      \<CommerceType\>COMMERCE_GOLD\</CommerceType\>
      \<iExtraCommerce\>5\</iExtraCommerce\>
  \</BuildingExtraCommerce\>
\</GlobalBuildingExtraCommerces\>

It seems like it couldn't give multiple commerce types, or it can but is extremely clunky.
It could look like bonus/tech commerces/yield in XML

\<GlobalBuildingExtraCommerces\>
    \<BuildingExtraCommerce\>
            \<BuildingType\>BUILDING_TAILOR_SHOP\</BuildingType\>
            \<Commerces\>
                  \<iCommerce\>5\</iCommerce\>
            \</Commerces>
      \</BuildingExtraCommerce\>
\</GlobalBuildingExtraCommerces\>

PR #1285, opened 2022-01-12
switched CvBuildingInfo::GlobalBuildingCommerceChange to IDValueMap

MattOttawa:
I havn't lookd at the xml yet.
closes: #1164


raxo2222:
Adjusted its XML, data is read correctly from XML.

PR #1287, opened 2022-01-13
added global define to set update rate

MattOttawa:
No idea what this does. It was just a test.
Delete this later if it's pointless or if nobody wants it.
I just wanna delete this branch.


MattOttawa:
Actually this speeds up/slows down graphics.

Toffer90:
I think it's how often the exe calls the python function called update(); that all UI classes have. So the refresh rate of UI.

Toffer90:
If it's the animation speed of units, which is another thing we can modify with an exe call like that, then having it as a global define would be risky, as that value shouldn't b messed with unless it is needed to solve some glitch by speeding up for a very short moment or something like that. But again I'm pretty sure this is the one for UI refresh rate.

PR #1300, opened 2022-01-18
Merge of animals XML GameText

SannitaSSJ:
This branch was created to re-order files related to animals (buildings, units, unsorted stuff, pedia), and if possible to fix entries here and there.


raxo2222:
You can't change infotypes themselves!

Only their TXT keys.

SannitaSSJ:
I thought I changed only the keys, not the infotypes!

raxo2222:
Well you didn't copy whole text string.
So BUILDING_GOVENORS_MENAGERIE got replaced with BUILDING_GOVERNOR_MENAGERIE for example

SannitaSSJ:
Ok, I'll undo that commit, and then re-try.

PR #1313, opened 2022-01-21
added const to CvGame::cyclePlotUnits

MattOttawa:


Toffer90:
Tested, doesn't CTD this time.

MattOttawa:
Thanks, I actually don't have a mouse.

PR #1330, opened 2022-02-04
Update Promotions_CIV4GameText.xml

paulcobbaut:
There is no c in exile.


raxo2222:
There are lots of typos in mod, so you can try to find more of them

PR #1331, opened 2022-02-05
Excile exile typo (untested)

paulcobbaut:
Changing 15 files to correct the "Excile-Exile" typo.

Some .cpp files are changed, but I do not have a development environment setup to compile C2C. So this is untested and may break...

Should I hold off such pull requests until I can compile C2C myself?

### DIFF
--- a/Assets/XML/Art/CIV4ArtDefines_Building.xml
+++ b/Assets/XML/Art/CIV4ArtDefines_Building.xml
@@ -14814,14 +14814,14 @@
 			<Button>Art/interface/buttons/buildings/exaverse.dds</Button>
 		</BuildingArtInfo>
 		<BuildingArtInfo>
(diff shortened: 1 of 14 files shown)


raxo2222:
If you set this mod to be gameplay mod (that is if you followed instructions to setting up it), then DLL will autocompile every time you do changes to cpp files.
Tools folder also have makedll files

Toffer90:
The PR itself does a compile test, and it compiled successfully.

The third one there is dll compilation test:
image

Toffer90:
Only problem is that this:
image
breaks save compatibility.

Toffer90:
> Only problem is that this: image breaks save compatibility.

Most practical solution is to simply not rename this one variable.

paulcobbaut:
Thanks for all the replies. I will update this later.

PR #1332, opened 2022-02-09
Excile typo take2

paulcobbaut:
Changing Excile into Exile everywhere,
except the .dds filename,
except the m_iExcileCount variable.

Putting SAVEBREAK before every occurrence of m_iExcileCount.

I hope this does not break anything, this is my first real pull request on C2C. I am open to suggestions, even critical ones.


flabbert:
thank you for this, we will push this in when we do our next save break, Ill notify here just before as well, so you can update any mergeconflicts at the time

PR #1350, opened 2022-03-18
added NEXT_TRAIT_CULTURE_REQ_PERCENT global define

MattOttawa:


raxo2222:
This only controls base, not steepness of it.

Next trait is 8x more expensive compared to previous one no matter what setting is.

PR #1362, opened 2022-04-10
made viewports the same size as the initial map when viewports are off

MattOttawa:
Already did this, then reverted. Raxo said it's better this way.
I forget where I'm at with this...

### DIFF
--- a/Sources/CvViewport.cpp
+++ b/Sources/CvViewport.cpp
@@ -77,7 +77,7 @@ void CvViewport::resizeForMap()
 		//	Force-load the main interface BUG module so we can get at the viewport BUG settings
 		Cy::call("CvAppInterface", "initBUG");
 	}
-
+#ifdef OLD_CODE
 	m_iXSize = GC.getENABLE_VIEWPORTS() ? GC.getVIEWPORT_SIZE_X() : m_pMap->getGridWidth();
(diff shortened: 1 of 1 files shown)


MattOttawa2:
Hmm... This branch fixes some weird bugs when switching to a smaller map, but it's still a bit odd.
The map never actually changes size, with or without this change.

Switching to a map that's bigger than the initial map crashes I think.
But if we can get smaller maps working good, then we can make the viewport the size of the biggest map.

Have you tried switching the map yet @Toffer90? I might need help.

If anyone wants to try, use debug DLL to enable.
Hit alt 0 for initial map.
Alt 1 is underground (no WB map yet, so doesn't really work).
Alt 2 is cislunar (just a black map).
Try alt 3 - alt 9.
Ctrl 0 - 6 are more maps.

Toffer90:
> Have you tried switching the map yet @Toffer90? I might need help.
> 
> If anyone wants to try, use debug DLL to enable. Hit alt 0 for initial map. Alt 1 is underground (no WB map yet, so doesn't really work). Alt 2 is cislunar (just a black map). Try alt 3 - alt 9. Ctrl 0 - 6 are more maps.

Thanks for reminding me, I tested once long ago when this was fresh, I may have more time and inclination to look deeper at this now than then. ^^

PR #1374, opened 2022-06-07
exposed 3 methods to python for flavors

MattOttawa:


MattOttawa:
My python test code can be deleted.

PR #1395, opened 2022-08-22
Domestic advisor: clicking on a city opens the city screen.

klara-zielinska:


dzfischer:
Can't say I'm familiar with this part of the code, but a quick check shows it does work as advertised. If Toffer objects, can revert, but seems like it's been a while since you've worked on this bit?

PR #1396, opened 2022-08-22
Unselect the selected unit or city on ESC

klara-zielinska:


dzfischer:
This prevents the main game menu from opening with escape, which isn't so great. It does unselect a unit that already has orders, but if the unit is active it just flashes the unit bar before reselecting it.
Expected behavior should probably be:
If unit is active or no units/cities selected, esc opens/closes the game menu
If unit has orders or city selected, esc closes.
If city has no build queue and there are no other active units, cities also lacking build queue, or pending events, esc might be disabled then as well to avoid the popup bouncing (but that'll also need to be implemented in other ways that one could exit the screen i.e. by clicking, pressing space, etc).

Toffer90:
> This prevents the main game menu from opening with escape, which isn't so great. It does unselect a unit that already has orders, but if the unit is active it just flashes the unit bar before reselecting it. Expected behavior should probably be: If unit is active or no units/cities selected, esc opens/closes the game menu If unit has orders or city selected, esc unselects them. If city has no build queue and there are no other active units, cities also lacking build queue, or pending events, esc might be disabled then as well to avoid the popup bouncing (but that'll also need to be implemented in other ways that one could exit the screen i.e. by clicking, pressing space, etc).

Ok, since the button W already does what this PR wants to achieve with esc, I think we can close it due to the bugs, which would be hard to do anything with as esc is a hotkey that the exe interacts with.

klara-zielinska:
> This prevents the main game menu from opening with escape, which isn't so great. It does unselect a unit that already has orders, but if the unit is active it just flashes the unit bar before reselecting it. Expected behavior should probably be: If unit is active or no units/cities selected, esc opens/closes the game menu If unit has orders or city selected, esc unselects them. If city has no build queue and there are no other active units, cities also lacking build queue, or pending events, esc might be disabled then as well to avoid the popup bouncing (but that'll also need to be implemented in other ways that one could exit the screen i.e. by clicking, pressing space, etc).

I deeply disagree. Opening the main menu is a functionality that is used rarely and you can always click the button. From the other hand when you get stuck with a selected unit that has no move, the first thing you try is pressing ESC.

The general intuition says that if you have a selected entity, then ESC button will drop the selection and exactly this is done. A unit flashes after hitting ESC, because is unselected, but if it still have moves, the engine chooses it again as the first unit that can move.

Toffer90:
> > This prevents the main game menu from opening with escape, which isn't so great. It does unselect a unit that already has orders, but if the unit is active it just flashes the unit bar before reselecting it. Expected behavior should probably be: If unit is active or no units/cities selected, esc opens/closes the game menu If unit has orders or city selected, esc unselects them. If city has no build queue and there are no other active units, cities also lacking build queue, or pending events, esc might be disabled then as well to avoid the popup bouncing (but that'll also need to be implemented in other ways that one could exit the screen i.e. by clicking, pressing space, etc).
> 
> I deeply disagree. Opening the main menu is a functionality that is used rarely and you can always click the button. From the other hand when you get stuck with a selected unit that has no move, the first thing you try is pressing ESC.
> 
> The general intuition says that if you have a selected entity, then ESC button will drop the selection and exactly this is done. A unit flashes after hitting ESC, because is unselected, but if it still have moves, the engine chooses it again as the first unit that can move.

I'll consider reopening this PR next weekend when I get time to look at it in game and evaluate it for myself.

dzfischer:
> I deeply disagree. Opening the main menu is a functionality that is used rarely and you can always click the button. From the other hand when you get stuck with a selected unit that has no move, the first thing you try is pressing ESC.
I guess it's a manner of personal preference; I often hit esc to open that menu instead of aiming the mouse to the corner, and hit enter or space to get the next unit if stuck at one without actions 🤷

klara-zielinska:
I have thought that ENTER ends the turn and SPACE changes the orders of the selected unit to idle. The first is bad if you still have other units to move. The second is bad if the selected unit has orders, like fortify or disease control, because the order will be canceled.

dzfischer:
> I have thought that ENTER ends the turn and SPACE changes the orders of the selected unit to idle. The first is bad if you still have other units to move. The second is bad if the selected unit has orders, like fortify or disease control, because the order will be canceled.

Enter ends the turn only if you have no other active units that can be selected; it's shift-enter that will force end the turn if there are still things to do. You are correct on space; I just often find myself using it because often the 'selected but no move' unit is one that I was just doing something with, and space ensures I remember about it next turn.
> Ok, since the button W already does what this PR wants to achieve with esc,

'Course, looks like we're both missing a keyboard shortcut though, will have to test this one 😅

Toffer90:
Ok, I'll commit a tweaked version of this feature, it will only unselect units if the selected unit has either red or grey status dot, i.e. unit cannot move or unit is not waiting for orders.

PR #1397, opened 2022-08-23
culture level option fix

### LINKED ISSUE #1394: CultureLevelInfo xml isn't parsed properly
`getCultureLevelInfo` and `getNumCultureLevelInfos` count all culture levels from the xml, not caring whether the realistic culture gameoption is on or not. As a result, there are multiple places in the dll that check this inconsistently. For example
Going from "Developing" to "Refined" in a standard game (skipping over Prominent) causes borders to grow twice, because level has grown by 2. Similar effects are likely all over.

Example workaround for updateCultureLevel:
[Bad fix on realistic-culture-tweaks branch, will need removal if done properly!](https://github.com/caveman2cosmos/Caveman2Cosmos/commit/7fa123a805a18f97fe4ca769d085b371411172df)

A proper fix should probably involve fixing the two base functions to appropriately count levels and get info, only for levels that are applicable due to game options. Flabbert has said:
"this is the time you create a new class that holds the culture calculations, create a new calculator there, and call that everywhere
you can even make it static"

Sadly, I have concluded I don't understand well enough to do myself 😓

MattOttawa:
Maybe this will fix #1394

### DIFF
--- a/Sources/CvCity.cpp
+++ b/Sources/CvCity.cpp
@@ -10839,6 +10839,8 @@ void CvCity::updateCultureLevel(bool bUpdatePlotGroups)
 	{
 		return;
 	}
+	uint32_t iCultureLevel = 0;
+
 	CvGameAI& GAME = GC.getGame();
 
 	if (!isOccupation()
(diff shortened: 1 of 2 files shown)


dzfischer:
Unfortunately I think the issue is a little more generic than that. If I'm reading this right, it'll set the culture level of the city to the "correct" level, but there are other parts of the code that do checks like "Is culture level one less than number of culture infos? If so, then it counts toward culture victory!" and such.
It does work as a spot fix, but there's a number of places that'll need similar spot fixing.
Could be done, but I think (?) updating the getNumCultureInfos might be a better solution? Or not, for some reason I do not know.

Issue #1080, opened 2021-07-19
Add fourth ring radius for cities at Megacities tech (equivalent of Metropolitian Administration)


raxo2222:
Main purpose of it would be ability to space out space colonies.
It wouldn't change much for Earthly cities - maybe except reaching deeper into ocean.

Space colonies would start with third ring radius - they all have autobuilding that is spawned when space city is settled, it can provide access to third ring.
At some point space cities have access to building, that expands them to 3rd ring - those can be changed to allow access to fourth ring.
This way you could place space cities every 6 tiles in big grid, and not have wasted tiles.

![Civ4BeyondSword_2021-07-19_17-11-39-54](https://user-images.githubusercontent.com/54320715/126188032-be9c1ba5-21c8-4b8a-88f6-b3a23a4e40d4.png)

flabbert:
this requires a few things to work, amongst other things, the plotmap needs to be updated with additional cityplot indexes, there are a couple of other things on the top of my head that is required as well. But should be doable

Issue #714, opened 2020-10-29
Realistic Culture Spread gameoption fails badly on conquered cities


dzfischer:
A conquered city will grow as if the option were off (thus reaching more tiles far sooner than it should), but only put one point of culture onto tiles incorrectly gained this way.

Kinda breaks gameplay - see more description: https://forums.civfanatics.com/threads/realistic-culture-spread-bug.677053/

Issue #246, opened 2019-11-16
IDW feature is unbalanced - its effects are rarely felt.


raxo2222:
Link to thread https://forums.civfanatics.com/threads/single-player-bugs-and-crashes-v39-download-after-july-20th-2019.647929/page-3#post-15524849

Not sure, if it is XML/Python/DLL issue.

PR #1400, opened 2022-08-26
Realistic culture tweaks

dzfischer:
Still needs some things.
Required:
- Fix issue #1394; either fix fetching culture levels and info for game options, or spotfix culture the rest of places they're called (~10 places). Possibly related to #99 ?
- ~~Fix issue with fixed borders not properly only grabbing some culture tiles on city conquest (part: https://github.com/caveman2cosmos/Caveman2Cosmos/commit/2caf3e630dce9c1f403e31f1e94856d2b0cfb66a#commitcomment-82273984 , see discord devtalk for pics of current state)~~
- Profile turntime, make sure haven't caused perf issues
  - recalculateCultureDistances should be spiral instead of while loop over grid, use/tweak getCityIndexPlot
  - ~~Check if there's places culture plotgroup is being needlessly updated~~
- ~~Add gamespeed modifier to cvPlot::doDecay()~~
- ~~Do an autoplay test, check haven't accidentally broken save compat~~

Nice to do:
- Review IDW (#246 ):
  - ~~Fixed border effect on pillage, combat etc~~
  - ~~Change float values to int math, help prevent oos?~~
  - ~~Review idw functionality. Absolute culture change or percentile?~~
  - Adjust from player feedback
- Re-implement reduced-culture-fighting-on-team (marked with TODO)
- Review Helsinki cathedral wonder, check for other culture-fiddly effects
- Update pedia (but can also wait for pedia concept branch)
- May as well do #1080 
- ~~Adjust city revolt further; culture modifier should maybe be something like (attacker percent)/(defender percent, min 1) instead of linear 1x - 2x. Account for fixed borders tho; this uncaps the upper bound, because the lower bound via garrison units is also uncapped. Makes it less trivial to prevent cities from culture flipping.~~

Both of these should already be fixed:
- #714 
- #241


dzfischer:
Currently included features:
- Realistic culture expands more slowly, consistently, and responds to immediate circumstances (i.e. you can get border growth when you chop forests/jungle, get tech that lets you use terrain easier, not just on new city culture level)
- More culture levels if on realistic culture
- Fixed bug with tiles flip-flopping between owned and unowned
- Fixed bug with conquered realistic culture cities expanding like nobodies business
- Fixed longstanding bug that gave extra border radius of cities lvl 4+ when realistic culture off
- Culture thresholds more regular: each level is 2x or 4x the previous cost, depending on if realistic culture is on
- Culture decays on all plots. The culture on a plot from a city reaches an equilibrium value fairly quickly related to the culture output of that city and distance from it, instead of stockpiling indefinitely. Some XML tags to control this (distance effect and decay both)
- No immediate change of plots on city conquest (decay and or IDW will switch control over time)
- Culture revolts have additional notifications telling you when one is possible, and again when you're close to achieving one.
- Fixed Borders now (consistently) makes the owner not lose control of a tile they already own until next player has twice (can change in xml) their culture. Also, so long as there's a military unit on an owned tile, will never flip.
- Fort tiles can flip from culture, but only if there are zero military units on it.

dzfischer:
Feature:
- When conquering (not razing) a city, gain updated knowledge of the terrain near the city itself (based on city workable radius), though not active line of sight
- If conquerer has fixed borders, tiles near a conquered city that they have a minimum cultural bound on holding will be gained.

dzfischer:
Feature:
- Culture revolt chance depends far more heavily upon culture ratio, instead of 2x boost at 100% foreign culture, is 100x boost at foreign culture (garrison revolt protection from units still has no negative cap).
- Discovered borked UI for garrison protection, needs fixing.

dzfischer:
Seeing how I haven't really gotten much response, I think I'll just merge this into master & release now; I've tested a fair bit and am pretty sure it's got no gamebreaking bugs.
Given the scope I could well be wrong in which case a quick revert will be in order, but oh well.
Will keep the branch open though, as there are still some things that could use doing relative to it (radius 4, perf, removing some of the hacky bandaids, etc).

PR #1408, opened 2022-09-28
Fix hunter promo event

MattOttawa:
Need to test but the event buttons in WB don't seem to work for me.


MattOttawa:
That don't work. Unlocks way too many promos.

PR #1410, opened 2022-10-13
switched culture distance loops to foreach 

MattOttawa:


flabbert:
this looks good to me on first glance, I am still unsure about the actual caching but should remove hte out of bounds bug we currently have.

Sources/CvCity.cpp
```diff
@@ -6521,61 +6509,49 @@ void CvCity::recalculateCultureDistances(int iMaxDistance) const
...
+int CvCity::calculateCultureDistance(const CvPlot* mainPlot, int iMaxDistance) const
 {
 	// This function is the meat of Realistic Culture, determines extra distance/cost for tiles.
 	// Each point of distance corresponds 1:1 with level of city culture to reach it.
 	PROFILE_FUNC();
 
-	// find the current plot and the distance to it
-	const CvPlot* mainPlot = plotXY(getX(), getY(), iDX, iDY);
-
+	FAssertMsg(plotDistance(getX(), getY(), mainPlot->getX(), mainPlot->getY()) <= iMaxDistance, "test"); // Matt: does the following check ever pass?
```
dzfischer, line 6551: Not sure precisely the functionality of plot()->rect(iMaxDistance, iMaxDistance) as it calls this function, but if it effectively makes a rectangle of radius X, this check exists to avoid having to calculate needlessly the corners of said rectangle (relative coord X, X say) because our actual city's radius can't be larger than X itself.

Sources/CvCity.cpp
```diff
@@ -6521,61 +6509,49 @@ void CvCity::recalculateCultureDistances(int iMaxDistance) const
...
+			if (m_aCultureDistances[plotX] != iNewValue)
+			{
+				m_aCultureDistances[plotX] = iNewValue;
+				bHasChanged = true;
 			}
-		}
-		numLoops++;
-		if (numLoops > 100)
-		{
-			bHasChanged = false;
```
dzfischer, line 6562: I had issues every now and then with endless loops; easy to accidentially introduce one when fiddling here, and knowing when it's in one vs just taking a slow turn is annoying.

Function really shouldn't be a while(update) loop, but instead something like:
- Spiral out from center
- On current examined tile, look at neighbors to see if there's a shorter distance, and apply to current tile if so
- Attempt to push to neighbors if can be done. If a neighbor was not previously max int and was pushed, add it to a list to come back to later
- Move to next tile in spiral, and continue process.
Once spiral loop is done, do same algo but going down the list, adding entries to bottom of list if tiles get push-updated.

While loop is functional though for now, and it's not obvious how much perf gain there is without profiling.

MattOttawa: I'll add it back, but you shouldn't need it unless something isn't working right.

Sources/CvCity.cpp
MattOttawa: Next line will never happen right?

if (plotDistance(getX(), getY(), mainPlot->getX(), mainPlot->getY()) > iMaxDistance) return MAX_INT;

dzfischer: I think it will frequently happen, unless I'm misunderstanding something.

Pretend iMaxDistance is 2 (regular big fat cross, 21 tiles). This loop to calculate the cultural tile distance from the city center will be called not on just the 21 tiles, but the full 25 tiles in the 5x5 rectangle. Given the plot distance of the corner tiles will be 3, greater than iMaxDistance of 2, we don't want to update/recalculate everything just to update those corners we won't need anyway, right?

Not much impact with max distance 2, but at max distance 10 with 20x20 tiles, given it basically loops thru the entire thing again everything when a single tile needs updating, removing the 50-some tiles seems like a prudent step.

Toffer90:
```cpp
// 4 | 4 | 3 | 3 | 3 | 4 | 4
// -------------------------
// 4 | 3 | 2 | 2 | 2 | 3 | 4
// -------------------------
// 3 | 2 | 1 | 1 | 1 | 2 | 3
// -------------------------
// 3 | 2 | 1 | 0 | 1 | 2 | 3
// -------------------------
// 3 | 2 | 1 | 1 | 1 | 2 | 3
// -------------------------
// 4 | 3 | 2 | 2 | 2 | 3 | 4
// -------------------------
// 4 | 4 | 3 | 3 | 3 | 4 | 4
//
// Returns the distance between plots according to the pattern above...
inline int plotDistance(int iX1, int iY1, int iX2, int iY2)```

flabbert:
I suggest that this will be merged, and that the changes needed to corners are done in a separate branch

this fixes a ctd that has been infinitely annoying

PR #1415, opened 2022-12-29
fixed crash

MattOttawa:


Sources/CvUnit.cpp
```diff
@@ -214,7 +214,10 @@ CvUnit::~CvUnit()
 		// Don't need to remove entity when the app is shutting down, or crash can occur
 		if (!gDLL->GetDone() && GC.IsGraphicsInitialized())
 		{
-			gDLL->getEntityIFace()->RemoveUnitFromBattle(this);
+			if (plot())
```
Toffer90, line 217: `if (plot())` will never be true here, as the unit has already set its coordinates to INVALID_PLOT_COORD before the deconstructor is called.

If it's the crash that SO reported, with the guerilla units conquering the English city, then that CTD has already been fixed. SO's crash at least happened immediately after `gDLL->getEntityIFace()->RemoveUnitFromBattle(this);` was called from our dll, and it happened because a unit that was not involved in a battlle got removed from the battle that was ongoing at the time (calling this function on a unit not in battle is normally harmless if no battle is going on).

Toffer90: Oh, yeah, my point being that albeit this might fix SO's crash it might cause other issues to never tell the exe to RemoveUnitFromBattle here in the unit deconstructor.

PR #1422, opened 2023-02-06
fix research overflow

MattOttawa:
I don't know if this branch is ready...
Maybe I should change only the save/load to long long when saves are broken, and merge this branch later.


MattOttawa:
I made a mistake somewhere in this branch.

flabbert:
did we fix this on last savebreak?

PR #1424, opened 2023-02-07
City culture plot cache

dzfischer:
Has an issue where other civs will display tiles as not being influenced by cities when they are in fact.

This is caused by the cache clearing between player turn and AI turn; same cause for the 'culture last turn' vs 'this turn' cache. Should be fixed before accepting; whether to only show for player's own culture or by making new cache, or changing where cache clears, not certain. Thoughts Toffer?


dzfischer:
Known issue: Odd behavior with towers/fort radius culture on high gamespeed on eq. culture, either not having appropriate radius or losing control turn after building. Normal speed or normal culture is fine though.
Will need to write better catch for this case in later svn ver.

PR #1438, opened 2023-03-12
Feature/combat and unit ai as services

flabbert:
This PR is to track the current changes between this and Master, So its easier to get an overview of the merge conflict, and see what, if anything can be done to get this into master, either wholesale or piecemeal


flabbert:
this is very outdated with current codebase unfortunately, will close PR, but leave branch for inspiration with the work that we are currently doing

PR #1439, opened 2023-03-15
Extra profiling

klara-zielinska:


Sources/CvBuildingInfo.h
```diff
@@ -568,6 +569,7 @@ class CvBuildingInfo
 	int m_iMaxStartEra;
 	TechTypes m_iObsoleteTech;
 	int m_iPrereqAndTech;
+	int m_iNoBonus;
```
MattOttawa, line 572: I'm a bit confused... maybe... but I think this branch accidentally reverts at least 1 commit.
https://github.com/caveman2cosmos/Caveman2Cosmos/commit/ebb6844342dfe9777a23797cd5c64ab396716aa0

klara-zielinska: Reverted. I am doing it once again. I must have merged master into the branch incorrectly.

PR #1442, opened 2023-03-29
Fixed a possible buffer overflow during culture decay

sp00n:
`getCulture(ePlayerX) * (1000 - decayPermille) / 1000` could apparently trigger a buffer overflow if the plot culture is around 2,147,483 even though the value is divided by 1000 again.
Reversing the order of operations to divide by 1000 first fixed this, but introduced another problem if the initial culture is lower than that, so we have to make an if else check. I've set the limit for reversing the order to 1 million.

Be aware I haven't tested the created DLL (it compiles fine) since I don't have a savegame for this version where this problem manifests, only for v43.
I have tested a similarly adjusted code for v43, but there have been significant changes to CvPlot.cpp since then, which are incompatible with each other, so this requires evaluation.

### DIFF
--- a/Sources/CvPlot.cpp
+++ b/Sources/CvPlot.cpp
@@ -10323,33 +10323,56 @@ void CvPlot::doCulture()
 	{
 		const int decayPermille = GC.getTILE_CULTURE_DECAY_PERCENT() * 1000 / GC.getGameSpeedInfo(GC.getGame().getGameSpeedType()).getSpeedPercent();
 
+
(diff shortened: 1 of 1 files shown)


dzfischer:
Toffer fixed issue with [https://github.com/caveman2cosmos/Caveman2Cosmos/commit/51d2971cd8d7b59f9625c37f43becb5ea18bc73d](https://github.com/caveman2cosmos/Caveman2Cosmos/commit/51d2971cd8d7b59f9625c37f43becb5ea18bc73d)

PR #1444, opened 2023-05-26
No-rng Battles Checkbox in BUG (C2C tab) options.

Leodim79:
As described and discussed here : https://forums.civfanatics.com/threads/quick-questions-quick-answers.647831/post-16459422


flabbert:
this is probably something that is still interesting @Toffer90 can you have a shifty here and see if it breaks stuff? or if we can put it in

PR #1448, opened 2023-08-06
Italian translation

Giuzy:
Only text translation for Italian


Toffer90:
Your first mini PR has been approved, we look forward to future PR's from you. ^^

PR #1454, opened 2023-09-03
Subcombat unit types and normal units (first part)

Giuzy:
As always, please let me know in case of problems. Thank you!

### DIFF
--- a/Assets/XML/GameText/Global_CIV4GameText.xml
+++ b/Assets/XML/GameText/Global_CIV4GameText.xml
@@ -86,14 +86,17 @@
 	<TEXT>
 		<Tag>INTERFACE_CITY_FOOD_GROW</Tag>
 		<English>%d1[ICON_FOOD] - %d2[ICON_EATENFOOD] = +%d3[ICON_FOOD]</English>
+		<Italian>%d1 [ICON_FOOD] - %d2 [ICON_EATENFOOD] = +%d3 [ICON_FOOD]</Italian> <!-- city screen left of the food bar -->
 	</TEXT>
 	<TEXT>
 		<Tag>INTERFACE_CITY_FOOD_SHRINK</Tag>
 		<English>%d1[ICON_FOOD] - %d2[ICON_EATENFOOD] = [COLOR_WARNING_TEXT]%d3[COLOR_REVERT][ICON_BAD_FOOD]</English>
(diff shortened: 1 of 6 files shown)


Assets/XML/GameText/Global_CIV4GameText.xml
```diff
@@ -22110,10 +22119,10 @@
 	<TEXT>
 		<Tag>TXT_KEY_RELIGION_FOUNDED_FIRST</Tag>
 		<English>[ICON_BULLET]Founded by First to Discover [LINK=%s1]%s2[\LINK]</English>
-		<French>[ICON_BULLET]Religion fondée par le premier à découvrir : [COLOR_TECH_TEXT]%s1_TechName[COLOR_REVERT]</French>
-		<German>[ICON_BULLET]Durch den ersten Entdecker der Technologie [COLOR_TECH_TEXT]%s1_TechName[COLOR_REVERT] gegründet</German>
-		<Italian>[ICON_BULLET]Fondato dalla prima civiltà che scopre [NUM1:il:la:i:le] [COLOR_TECH_TEXT]%s1_TechName[COLOR_REVERT]</Italian>
-		<Spanish>[ICON_BULLET]Fundado por el primero en descubrir [NUM1:el:la:los:las] [COLOR_TECH_TEXT]%s1_TechName[COLOR_REVERT]</Spanish>
+		<French>[ICON_BULLET]Religion fondée par le premier à découvrir : [COLOR_TECH_TEXT]%s2[COLOR_REVERT]</French>
```
Giuzy, line 22122: This and the following 3 were changed on purpose. English was the only one having the text appearing as it has another tag, I've changed the wrong ones for Italian, French, German and Spanish.

Giuzy: BUG5DEU
BUG5FRE
BUG5SPA
BUG5

Giuzy: Another thing, I noticed in all languages we have 2 units with the same name: Camel Archer. One is a normal unit, the other a specific culture one. Normally it's not very good to have this situation, so I modified the second one for Italian into "Arab camel archer". Just to let you know, if you want to avoid the duplicate too. Ciao.

Giuzy: DoubleArcher

Toffer90: You should add the [LINK=%s1]...[\LINK] around the %s2 where it is missing here while you're at it.
The link will add colour so the color part can be replaced by the link part.

Giuzy: Done, you're right didn't notice that. Thanks for checking!

PR #1460, opened 2023-10-10
Added obsidian resource to mine

ChristophGra:
The Mine not providing resources while the stone tools workshop, which upgrades into a mine does, seemed like a bug. So this small edit should fix it. Tested it on my local version and worked perfectly fine there.

I copied the bonus_obsidian from the quarry, but adjusted the idiscoverRand and IDepletionRand to be in line with other mine resources.


ChristophGra:
Thinking more about it, obsidian provided by a mine doesn't really make sense, as it is already provided by a quarry.

PR #1463, opened 2023-11-25
Fixes for Russian localization

Qwest8K:


flabbert:
@Qwest8K have you playtested this?

Qwest8K:
@flabbert yes.
You found errors, which I missed?

flabbert:
no, simply checking before merge

PR #1473, opened 2024-03-09
Feature/event rework and fork pull request test

SimoCvijic0:


Assets/Python/EntryPoints/CvRandomEventInterface.py
```diff
@@ -6343,6 +6343,24 @@ def getHelpVolcanoExtinction(argsList):
...
+
+	for i in range(GC.getNumBuildingInfos()):
+		if isLimitedWonder(i) or not CyCity.hasBuilding(i) or CyCity.isFreeBuilding(i):
+			continue
+		info = GC.getBuildingInfo(i)
+		if info.getProductionCost() < 1 or info.isNukeImmune() or info.isAutoBuild():
+			continue
+	        iBurnBuilding = i
+
+	if iBurnBuilding != -1:
```
Toffer90, line 6359: iBurnBuilding needs to be set to be -1 before the for loop so it won't fritz out for cities with no valid buildings to burn.

Assets/XML/GameText/Events_CIV4GameText.xml
```diff
@@ -861,6 +861,14 @@
 		<Russian>Сообщают, что из-за мощных ураганов сильно пострадали рыбаки в державе %s1_civ_adjective.</Russian>
 		<Spanish>Un fuerte temporal ha interrumpido la actividad pesquera en el imperio %s1_civ_adjective.</Spanish>
 	</TEXT>
+    <TEXT>
+        <Tag>TXT_KEY_EVENTTRIGGER_MASSIVE_FOREST_WILDFIRE</Tag>
+        <English>Due to recent droughts a massive wildfire has started near the %s1_civ_adjective city of %s2_city.</English>
+        <French></French>
+        <German></German>
+        <Italian></Italian>
+        <Spanish></Spanish>
```
Toffer90, line 870: Just delete the empty language lines, here. Doesn't matter much either way, just a subjective preference from me.

Toffer90:
Seems good, have you experienced all these events in game and seen that they work as expected?

SimoCvijic0:
I will fix this then continue with other events in a separate branch in the fork

Assets/Python/EntryPoints/CvRandomEventInterface.py
Toffer90: Another thing here that should be improved is to add all the valid buildings to a list, and then pick a random one of them from the list to burn. Right now it will always burn the building with the highest index value in the city which will mean buildings will burn down in a specific order, that lower index buildings will be very well insulated and rarely burn because of it.

Toffer90: > Another thing here that should be improved is to add all the valid buildings to a list, and then pick a random one of them from the list to burn. Right now it will always burn the building with the highest index value in the city which will mean buildings will burn down in a specific order, that lower index buildings will be very well insulated and rarely burn because of it.

An example of what I mean:
```py
def applyMercenariesAncient2(argsList):
	data = argsList[1]
	listPlots = []
	MAP = GC.getMap()
	for i in xrange(MAP.numPlots()):
		plot = MAP.plotByIndex(i)
		if (plot.getOwner() == -1 and not plot.isWater() and not plot.isImpassable()
		and plot.area().getCitiesPerPlayer(data.ePlayer) > 0
		and plot.isAdjacentPlayer(data.ePlayer, True)
		):
			listPlots.append(i)

	if not listPlots:
		return

	plot = MAP.plotByIndex(listPlots[GAME.getSorenRandNum(len(listPlots), "Mercenary event location")])
```

SimoCvijic0: hmm very interesting thing i noticed as well in other burning events is that it always burns the same houses. I guess this was the cause.

PR #1489, opened 2024-04-03
Negative happiness being applied to cities while merging

ambro-mlody:
On turn 0 AI merges units that it start's with. If they build city at the starting plot in the same turn, negative happiness is applied to city, which results in starting city having at the highest difficulty -6 happiness. It is caused by kill delay.


ambro-mlody:
Screenshot_25
Capturet
cosmeticassert
lol

Toffer90:
I'll look a bit into this today or tomorrow to make sure this address the problem adequately at the source/root.

PR #1491, opened 2024-04-07
added InitGraphics to switchMap()

MattOttawa2:
This actually changes map size, but it makes the fog buggy.
Switching to bigger maps still crashes.


flabbert:
could you describe the bug?

MattOttawa2:
I figured it out.

PR #1495, opened 2024-04-19
Temp fix to workers not doing their job

ambro-mlody:
Clearing team border cache

### DIFF
--- a/Sources/CvPlayer.cpp
+++ b/Sources/CvPlayer.cpp
@@ -12235,6 +12235,7 @@ void CvPlayer::setTurnActive(bool bNewValue, bool bDoTurn)
 		gDLL->getInterfaceIFace()->setDirty(Score_DIRTY_BIT, true);
 
 		GC.getMap().invalidateIsActivePlayerNoDangerCache();
+		GC.getMap().invalidateIsTeamBorderCache(getTeam());
 	}
 }
 


Toffer90:
Are you sure this change fixed the bug of AI gatherers not doing their job?

I can't see how this changes anything, all that cache does is to mark plots that are within 1-2 plots of the border of a nation you are at war with. The cache is invalidated correctly when culture ownership change and when war and peace are declared.
The only place the cache is used is where it will correctly predict if the plot is within 1-2 plots of an enemy border. The function that use the cache only use it to skip a lot of complicated routines looking for reasons for danger on the plot, one of them being if it the plot is close to an enemy border. The cache will be reset, the moment it is used, to the value it had before it was invalidated at the start of the turn, and the function that use it will return the same value and do the same job as if it had not been invalidated in the first place. So invalidating it when it is not invalid just waste process time.

flabbert:
We have definitely seen evidence of improved work behaviour after this cache gets reset, is it the root cause? no, thats not sure, what alberts8 wrote in the bug thread on discord probably is a more comprehensive fix

ambro-mlody:
Basically what flabbert said. It's more a temp fix until we find actual cause. But it is definitely doing something. There is probably something wrong with setting cache. I'm not sure about setting it to both teams. Just because some plotX is close to an enemy border, that does not necessarily mean, that plot on that border, owned by an enemy, is close to our border. plotX may be neutral.

Toffer90:
> Basically what flabbert said. It's more a temp fix until we find actual cause. But it is definitely doing something. There is probably something wrong with setting cache. I'm not sure about setting it to both teams. Just because some plotX is close to an enemy border, that does not necessarily mean, that plot on that border, owned by an enemy, is close to our border. plotX may be neutral.

` Just because some plotX is close to an enemy border, that does not necessarily mean, that plot on that border, owned by an enemy, is close to our border. plotX may be neutral.`

Yeah, this was the main culprit, but there were more less serious glitches in it.

PR #1498, opened 2024-04-26
finished fixing the missions that move units to a different map

MattOttawa2:


Toffer90:
image

MattOttawa2:
Yes.
Here's another from you I think.
Screenshot 2024-04-27 202922

Toffer90:
> Yes. Here's another from you I think. Screenshot 2024-04-27 202922

Yeah, I fixed that one right before noticing you about your warning, as I saw both warnings.

PR #1528, opened 2024-07-24
Fix first time switch into a multimap causing turn count and year to reset

ToasterKoishi:


Assets/Python/pyWB/CvWBDesc.py
```diff
@@ -1907,8 +1907,10 @@ def read(self, fileName):
 		return 0
 
 	# after reading setup the map
-	def applyMap(self):
-		self.gameDesc.apply()
+	def applyMap(self, shouldApplyYearTurn = True):
+		if shouldApplyYearTurn:
```
MattOttawa2, line 1911: Pretty good. shouldApplyYearTurn could be GAME.getCurrentMap() == MapTypes.MAP_EARTH:

PR #1529, opened 2024-07-24
Fix moving units from Earth to Cislunar placing them in non-Orbit plots; Fix space unit mission rewards not moving to capital properly

ToasterKoishi:
Not sure how easy it is to be able to instantly TP a unit between multimaps, as trying to do a `switchMap` during `doOutcome` (regardless of if the switch was in the XML/Python or in actual C++) would crash the game. However to properly create a unit on Earth or get the capital's coords, the current map does need to be Earth... too much random stuff everywhere just assumes to use GC.getMap instead of being passed a map to operate on


MattOttawa2:
So this fixes your problem? I can merge it?
I have no idea how satellites are supposed to work.

If you want units to switch maps instantly that shouldn't be hard to do.

ToasterKoishi:
It would be best if I could switch maps instantly for some units, as that would better mirror the behavior that space maps currently has. Could you show me how to do that?

MattOttawa2:
I'm not sure how off the top of my head, but that's the way it was originally.
I'll take a look.

PR #1546, opened 2024-08-13
Paddle Steamer Movement Fix

The360MlgNoscoper:
The Paddle Steamer is a steam-powered ship. Let it ignore Terrain movement costs. Otherwise it's slower than any wooden ship with the Maneuvering 1 promotion.

### DIFF
--- a/Assets/XML/Units/U_Sea_CIV4UnitInfos.xml
+++ b/Assets/XML/Units/U_Sea_CIV4UnitInfos.xml
@@ -729,6 +729,7 @@
 			<bMilitaryProduction>1</bMilitaryProduction>
 			<bPillage>1</bPillage>
 			<bIgnoreBuildingDefense>1</bIgnoreBuildingDefense>
+			<bIgnoreTerrainCost>1</bIgnoreTerrainCost>
 			<bMechanized>1</bMechanized>
 			<bRenderBelowWater>1</bRenderBelowWater>
 			<UnitUpgrades>


Thunderbrd:
Although it gets a little rework on the unit redesign plan, it really is supposed to be slower than all other ships around it for 2 reasons.  1 - sail ships are actually very fast in this era and although they aren't subject to being so concerned about staying closer to shore, they are also fairly slow in comparison to peak sailing speeds. 2 - This lets us reset the ship speed for the industrial on at something more strategic without the game requiring ship speeds to accelerate to the point of madness for a turnbased game which makes any kind of actual strategic maneuvering impossible to depict navally.  There are still challenges here but at least for now, this was part of a set of solutions.

The360MlgNoscoper:
> Although it gets a little rework on the unit redesign plan, it really is supposed to be slower than all other ships around it for 2 reasons. 1 - sail ships are actually very fast in this era and although they aren't subject to being so concerned about staying closer to shore, they are also fairly slow in comparison to peak sailing speeds. 2 - This lets us reset the ship speed for the industrial on at something more strategic without the game requiring ship speeds to accelerate to the point of madness for a turnbased game which makes any kind of actual strategic maneuvering impossible to depict navally. There are still challenges here but at least for now, this was part of a set of solutions.

Slower than the Ironclad? The Paddle Steamer moves half as fast as any other naval unit you have had for the past 3 eras. Why this targeted hate towards specifically the paddle steamer? You know it actually has sails, right? It's half the speed of the galleon, the ship it's supposed to be an upgrade of. It's slower than any triremes you would still have around by this point. It sucks from a game design standpoint. And a realism standpoint. This is an emergency fix. The paddle steamer is the main transport unit for almost the entirety of the industrial age. There is no reason for it to be slower than ships from the classical age.

Forget the unit redesign for a second, this is actually a problem right now. Ask anyone who has actually played the game to the industrial age, they would tell you how silly it is for this unit specifically to not ignore terrain movement costs, which _every other steam ship does_. And it's one of the most important ones too.

The360MlgNoscoper:
Had a discussion with Thunderbrd on the Discord about this, he won't argue if this is implemented. This is only until the unit review is finished.

PR #1631, opened 2025-06-27
Civics balance update from Joseph

MattOttawa:


SimoCvijic0:
Is this mergable?

MattOttawa:
I think so.
We can revert if it's not exactly what Joseph wants.

SimoCvijic0:
And who is Joseph? Is this from the forum? ok ill merge in 5 min.

PR #1647, opened 2025-08-24
AI Fix : Limitation of See_Invisible and Property Control Units

Calvitix:
**1st Fix : See_Invisible**    
modifs in CvCityAI::AI_establishSeeInvisibleCoverage()

-  Reduce the reactivity if SEE_INVISIBLE needed (Unit is ordered only by 30% of time)
    => As that check is done every turn for every visibility

-     All Defensive/Counter Units in Town are evaluated to check the Visibility Intensity (not only the SEE_INVISIBLE in CvPlot::CountSeeInvisibleActive
    => a LE or Dog that is assigned to City_Defense or else (Not See_Invisibility), if it has Visibility promotion, will be count as active, and no new unit will be ordered.

-     Promotion Weight Value of VisibilityChange for SEE_INVISIBLE x2 and for VisibilityChangeRange x10 (value can be adjusted later). To ensure that the VisibilityChange promotions are chosen  (in CvPlayerAI::AI_unitValue, CvPlayerAI::AI_unitCombatValue and CvPlayerAI::AI_promotionValue)

**2nd Fix : Property Control Units :** 
Modifs in  CvCityAI::AI_choosePropertyControlUnit
- Fix to reduce the number of property control units that are created : princip is to approve the order of prop.control unit or building only if the value is going the wrong way. As soon as for example crime is lowering, the AI shouldn't do anything, Even if crime is still high, as it Will be Better in the futur.

**Not specific to Fix :**

-     Introduce more RNG (50+10% Value) for promotion Evaluation (was 25) => to encourage diversity in promotions.

-     Add a 4th level of log, for Log that generate a lot of trace. If _DEBUG, Level of Log is set to 4.

-     Add a lot of AIEvaluation Logs (Promotion Eval, See_Invisiblecoverage,...  in CvCityAI::AI_bestUnitAI, ...

-     Add a criteriaToString description to help log

+  Prepare some other improvement by add property param in AI_hasBeneficialPropertyEffectForCity


Sources/CvGame.cpp
```diff
@@ -9740,7 +9740,7 @@ void CvGame::changeHighToLowCounter(int iChange)
 void CvGame::doFinalFive()
 {
 	PROFILE_EXTRA_FUNC();
-	if (!isGameMultiPlayer() && isOption(GAMEOPTION_CHALLENGE_CUT_LOSERS) && countCivPlayersAlive() > 5)
+	if (!isGameMultiPlayer() && isOption(GAMEOPTION_CHALLENGE_CUT_LOSERS) && countCivPlayersAlive() > 10) //Calvitix TODO restore to 5
```
ivotentije, line 9743: you forgot to restore this

Calvitix: Good catch. How can I modify it. Closing and pulling a new request ?

